PR #1, opened 2016-06-29
Adding babel support and hooking into build system

ecwillis:


app.js
```diff
@@ -10,16 +10,22 @@ const config = require('confi')({
...
 mkdirp.sync(path.join(config.CWD, '.dist'));
 
 const watchedFiles = [
   path.join(__dirname, 'styles', '**/*.css'),
+  path.join(__dirname, 'scripts', '**/*.js'),
 ];
 const stylesheets = Object.keys(config.stylesheets);
+const scripts = Object.keys(config.scripts);
 
 const watcher = chokidar.watch(watchedFiles, {
```
jgallen23: should we set up separate watchers for css and js?

app.js
```diff
@@ -10,16 +10,22 @@ const config = require('confi')({
...
+const jsProcessor = require('./tasks/script.js');
+
 const mkdirp = require('mkdirp');
 const debounce = require('lodash.debounce');
 
 mkdirp.sync(path.join(config.CWD, '.dist'));
 
 const watchedFiles = [
   path.join(__dirname, 'styles', '**/*.css'),
+  path.join(__dirname, 'scripts', '**/*.js'),
```
jgallen23: we should probably watch the client kit directory for yaml changes too, right?

ecwillis: Yeah - probably on both styles and scripts?

app.js
ecwillis: We probably should. @jgallen23

scripts/default.js
```diff
@@ -0,0 +1,11 @@
+'use strict';
```
jgallen23: what is this?

jgallen23: maybe move to an example directory or somethin?

app.js
dawnerd: unrelated to babel but on helpers as well

scripts/default.js
ecwillis: Yeah. just something to play with... Do we have anything to start with script wise?

dawnerd: domodule?

dawnerd:
Looks good to me

PR #2, opened 2016-07-01
Stylelint configuration suggestions

mturnwall:


.stylelintrc.json
```diff
@@ -16,20 +16,25 @@
     "function-name-case": "lower",
     "function-parentheses-newline-inside": "always-multi-line",
     "function-parentheses-space-inside": "never",
-    "function-url-data-uris": "never",
+    "function-url-data-uris": null,
```
mturnwall: Sometimes you might need to use a data URI

.stylelintrc.json
```diff
@@ -16,20 +16,25 @@
...
     "function-url-quotes": "always",
     "function-whitespace-after": "always",
-    "number-leading-zero": "never",
+    "number-leading-zero": "always",
```
mturnwall: Having the zero helps to avoid mistakes when someone doesn't notice the point, `.1` vs `0.1`. Also helps to convey that the author truly meant `0.1` and `.1` isn't a typo.

.stylelintrc.json
```diff
@@ -16,20 +16,25 @@
...
     "number-max-precision": 10,
+    "number-no-trailing-zeros": true,
     "string-no-newline": true,
     "string-quotes": "single",
     "length-zero-no-unit": true,
-    "time-no-imperceptible": true,
+    "time-no-imperceptible": [true, {
```
mturnwall: Probably will never need a time faster than 100ms but just in case I made this a warning to call attention to when someone does go under 100ms. It's a way of saying, "Are you sure you want this timing?"

.stylelintrc.json
```diff
@@ -44,6 +49,7 @@
     "declaration-colon-newline-after": "always-multi-line",
     "declaration-colon-space-after": "always",
     "declaration-colon-space-before": "never",
+    "declaration-no-important": true,
```
mturnwall: Nothing is ever that `!important`. If you need `!important` then more than likely you need to refactor your HTML. For the rare case of needing to override a 3rd party library rule this stylelint rule can be disabled inline.

.stylelintrc.json
```diff
@@ -100,6 +106,7 @@
     "comment-whitespace-inside": "always",
     "indentation": 2,
     "max-empty-lines": 1,
+    "max-nesting-depth": 4,
```
mturnwall: Avoid ending up in limbo.

.stylelintrc.json
```diff
@@ -16,20 +16,25 @@
...
-    "time-no-imperceptible": true,
+    "time-no-imperceptible": [true, {
+      "severity": "warning"
+    }],
     "unit-blacklist": ["em", "mm", "q", "cm", "in", "pt", "pc", "ex", "ch"],
     "unit-case": "lower",
     "unit-no-unknown": true,
     "value-keyword-case": "lower",
-    "value-no-vendor-prefix": true,
+    "value-no-vendor-prefix": [true, {
```
mturnwall: Made this a warning for the rare instances when the prefix is needed.

dawnerd:
👍

PR #3, opened 2016-07-02
Button mixin and documentation update

mturnwall:
At first I was trying to figure out how to make it more configurable and have hover states, but I decided to keep it as simple as possible for now.

Also added some quick documentation on how to use mixins since there was an issue with compiling. I documented the solution so no one else has the same issue.

### DIFF
--- a/readme.md
+++ b/readme.md
@@ -9,17 +9,4 @@ When working on clientkit make sure you have the following tools installed:
 
 Project has been setup to be run and debug from inside VS Code. Alternately you can run `npm start`.
 
-### Using Mixins
-
-When calling a mixin the first parameter is the name of the mixin and the values you want to pass to the mixin follow. You will need to use a comma to separate the values passed to the
-mixin but do not have a comma following the name of the mixin. Here is an example:
-
-```css
-.btn-primary {
-  @mixin button var(--color-accent), var(--color-link);
-  width: 300px;
-  padding: 10px 0;
-}
-```
-
-Notice that there is no comma following `button` but there is for the rest of the parameters.
+### [Mixins]
--- a/readme.md
+++ b/readme.md
@@ -8,3 +8,18 @@ When working on clientkit make sure you have the following tools installed:
    - Nodenv or similar (should match the version in .node-version)
 
 Project has been setup to be run and debug from inside VS Code. Alternately you can run `npm start`.
+
+### Using Mixins
+
+When calling a mixin the first parameter is the name of the mixin and the values you want to pass to the mixin follow. You will need to use a comma to separate the values passed to the
+mixin but do not have a comma following the name of the mixin. Here is an example:
+
+```css
+.btn-primary {
+  @mixin button var(--color-accent), var(--color-link);
+  width: 300px;
+  padding: 10px 0;
+}
+```
+
+Notice that there is no comma following `button` but there is for the rest of the parameters.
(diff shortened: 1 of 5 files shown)


styles/mixins/button.js
```diff
@@ -0,0 +1,10 @@
+module.exports = function () {
+  return function (mixin, bgColor, color) {
```
jgallen23: we also need a outline/hollow button. should this be part of this mixin or separate?  I guess we could just create a mixin that calls this mixing with a  border-color and then set bgColor to transparent.  thoughts?

readme.md
```diff
@@ -8,3 +8,18 @@ When working on clientkit make sure you have the following tools installed:
    - Nodenv or similar (should match the version in .node-version)
 
 Project has been setup to be run and debug from inside VS Code. Alternately you can run `npm start`.
+
+### Using Mixins
```
jgallen23: lets move this into a docs/mixins.md

styles/helpers/buttons.css
```diff
@@ -0,0 +1,5 @@
+.btn-primary {
+  @mixin button var(--color-accent), var(--color-link);
+  width: 300px;
+  padding: 10px 0;
```
jgallen23: I think padding should be part of mixin, you?

styles/mixins/button.js
```diff
@@ -0,0 +1,10 @@
...
+    return {
+      'background-color': bgColor,
+      color,
+      display: 'inline-block',
+      'line-height': 1
```
jgallen23: don't we need text-align: center too?

jgallen23: doesn't line height need to be the height of the button?

jgallen23: probably want to pass in height and then default it to 25px or so

styles/helpers/buttons.css
chrisopedia: Should we be forcing a width?

jgallen23: nope.  I think we should have left/right padding instead

styles/mixins/button.js
chrisopedia: Why not do padding on top and bottom. Then we can get different sizes

jgallen23: yeah, that works for me.  we probably also need a reset -webkit-appearance too, right?

styles/helpers/buttons.css
chrisopedia: Might be nice to have the ability to pass in a `max-width`

styles/mixins/button.js
chrisopedia: I know it's gross, but I'd only do that to an attributed selected input element.

```
input[type="button"],
Input[type="submit"]
```

mturnwall:
> webkit-appearance

Chris is right that that property should only be applied to button elements. Using attribute selectors is just fine.

> doesn't line height need to be the height of the button?

Setting the `line-height` to one helps to remove the differences on how browsers apply it. Using padding on top and bottom gives better consistency for sizing.

> we also need a outline/hollow button

Wouldn't the helper class take care of this? It would just be a class that calls the button mixin with a background color of transparent.

I made the mixin simple with the idea of helper classes adding the specifics.

> Should we be forcing a width?

This one I'm not sure about. As part of the design language will buttons always be the same size? Should the mixin be setting the width or should the site's theme set it? Will width's be changing at breakpoints? For example on desktop next and cancel buttons are next to each other but on mobile they span the full width stacked.

I added a `text-align` center since I don't think I've ever seen a button whose text wasn't centered aligned 😄

dawnerd:
@mturnwall can you merge in master and fix whatever conflict there is?

PR #5, opened 2016-07-09
Create: initial pass at tabs (css only)

chrisopedia:
Included is an example with the required HTML structure.

### DIFF
--- a/styles/helpers/tabs.css
+++ b/styles/helpers/tabs.css
@@ -1,0 +1,1 @@
+@mixin tabs 5, 6px;
--- a/styles/mixins/tabs.js
+++ b/styles/mixins/tabs.js
@@ -1,0 +1,65 @@
+module.exports = function () {
+  return function (mixin, number, borderRadius) {
+    var styles = {
+      '.tabs': {
+        display: 'flex',
+        'flex-wrap': 'wrap',
+        'max-width': '100%',
+        position: 'relative',
+        'list-style': 'none',
+        'border-radius': borderRadius,
+      },
+      '.tab': {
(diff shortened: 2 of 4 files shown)


styles/helpers/tabs.css
```diff
@@ -0,0 +1 @@
+@mixin tabs 5, 6px;
```
chrisopedia: Left in the border-radius section of this because its quite possible that we'd want rounded corners and the logic that was used within the CSS is complex enough that I didn't want to lose it.

styles/mixins/tabs.js
```diff
@@ -0,0 +1,65 @@
...
+          'user-select': 'none',
+        },
+        '&-content': {
+          position: 'absolute',
+          left: 0,
+          'z-index': '-1',
+          width: '100%',
+          'border-radius': borderRadius,
+          opacity: 0,
+          transform: 'translateY(-3px)'
```
dawnerd: whats this relative to?

chrisopedia: What do you mean?  This was copied from the example site that Greg had recommended and I stripped out all styling.  There is a series of checkboxes that have it reset to 0 and it appears this helps position the content below the tabs.  Also might just be to get hardware acceleration for the transitions.

dawnerd: Huh, -3px seems oddly specific though. If it looks fine and doesn't affect styles then 👍

chrisopedia: It does.  Take a look at the example.  We could remove it and see if there is issues with it as well.  Thoughts?

PR #6, opened 2016-07-10
Button mixin

mturnwall:
The button mixin is now used to create button variations.

You get output like this now. We can tweak the output as needed.

```css
.btn-primary {
  background-color: rgba(75, 157, 249, 1);
  color: orange;
}

.btn-primary:hover {
  background-color: red;
}

.btn-primary:active, .btn-primary.active {
  background-color: red;
}

.btn-primary:focus {
  background-color: red;
}
```

### DIFF
--- a/conf/default-colors.yaml
+++ b/conf/default-colors.yaml
@@ -2,6 +2,7 @@
 color:
   primary: '#333'
   secondary: '#666'
+  white: '#fff'
   accent: orange
   background: rgb(200,200,200)
   link: '{{color.primary}}'
@@ -9,6 +10,7 @@ color:
   input-background: '#fff'
   input-border: '#333'
   button-primary-bg: 'rgba(75, 157, 249, 1)'
-  button-primary-color: '{{color.accent}}'
+  button-primary-color: '{{color.white}}'
   button-secondary-bg: '{{color.primary}}'
-  button-secondary-color: '{{color.secondary}}'
+  button-secondary-color: '{{color.white}}'
+  button-primary-outline-color: '{{color["button-primary-color"]}}'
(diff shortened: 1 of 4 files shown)


styles/mixins/button.js
```diff
@@ -1,11 +1,15 @@
 module.exports = function () {
   return function (mixin, bgColor, color) {
-    return {
-      'background-color': bgColor,
-      color,
-      display: 'inline-block',
-      'line-height': 1,
-      'text-align': 'center'
+    const activeBgColor = 'red';
```
jgallen23: eh? red?

conf/default-colors.yaml
```diff
@@ -8,3 +8,7 @@ color:
   link-hover: '{{color.secondary}}'
   input-background: '#fff'
   input-border: '#333'
+  button-primary-bg: 'rgba(75, 157, 249, 1)'
```
jgallen23: this should be the accent

conf/default-colors.yaml
```diff
@@ -8,3 +8,7 @@ color:
...
+  button-primary-color: '{{color.accent}}'
```
jgallen23: this should be white (typically)

styles/mixins/button.js
mturnwall: I didn't know what colors to use so I just picked colors to test with.

jgallen23: we should never have colors defined in the mixin. should pull in from config or passed in

mturnwall: It's just a placeholder for testing. I mentioned this in the room but not sure anyone saw it. I'd like to add https://www.npmjs.com/package/postcss-colour-functions. We can use it for the hover effects on buttons. So `activeBgColor` would be `activeBgColor = darken(bgColor)`

dawnerd: I think hover colors need to be defined in colors.yaml

jgallen23: If we add support for those functions,we could still do darker and lighter in yaml

PR #9, opened 2016-07-15
Mixin refactor

jgallen23:
main reason for this is to be more explicit with the helper classes that are created.  There was no real way to leverage the mixins for typography and spacing, so I made them single use mixins and then explicitly added the helper classes and called the mixin for each. It's a bit of duplicate code, but will give us the ability to use font-size and spacing in other places.


jgallen23:
ok, a lot has changed here.  went back to dynamically generating helper classes so the yaml has more control, but also added support to be able to call one off mixins like `@mixin font-style heading-1`

jgallen23:
also added support for spacing and font styles do have a default style and then each breakpoint can override that.  we might need to think how max/min widths play into this.

lib/breakpoint-helper.js
```diff
@@ -0,0 +1,22 @@
+'use strict';
+
+module.exports = function(mixinStyles, breakpoints) {
+  console.log(mixinStyles, breakpoints);
```
dawnerd: Probably don't want to keep this here

PR #12, opened 2016-07-18
Update; fix the examples to point to correct CSS

chrisopedia:

The examples directory was changed to add another depth of folder, and thus the CSS was broken since it wasn't changed to reflect.

- update buttons example
- update tabs example

### DIFF
--- a/examples/tabs/index.html
+++ b/examples/tabs/index.html
@@ -6,7 +6,7 @@
     <meta http-equiv="x-ua-compatible" content="ie=edge,chrome=1" />
     <!-- iphone meta tag (width=device-width letterboxes on iphone 5) -->
     <meta name="viewport" content="width=device-width, initial-scale=1" />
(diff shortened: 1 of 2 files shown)


examples/tabs/index.html
```diff
@@ -6,7 +6,7 @@
     <meta http-equiv="x-ua-compatible" content="ie=edge,chrome=1" />
     <!-- iphone meta tag (width=device-width letterboxes on iphone 5) -->
     <meta name="viewport" content="width=device-width, initial-scale=1" />
-    <link rel="stylesheet" href="../.dist/tabs.css" />
+    <link rel="stylesheet" href="../../.dist/tabs.css" />
```
jgallen23: tabs is in examples/.dist, so this is correct

jgallen23:
this isn't needed if you run example

PR #14, opened 2016-07-19
Create; list-unstyled and list-inline

chrisopedia:

1. provides list mixins for both unstyled and inline
2. provides helper classes for both as well

### DIFF
--- a/styles/mixins/list-inline.js
+++ b/styles/mixins/list-inline.js
@@ -1,14 +1,15 @@
 module.exports = function () {
-  return function (mixin) {
+  return function (mixin, spacing) {
+    var spacing = spacing || '5px';
     return {
-      'margin-left': '-5px',
+      'margin-left': '-' + spacing,
       'padding-left': 0,
       'list-style': 'none',
 
       '& > li': {
         'display': 'inline-block',
-        'padding-right': '5px',
-        'padding-left': '5px'
+        'padding-right': spacing,
+        'padding-left': spacing
       }
     };
   };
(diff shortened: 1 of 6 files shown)


styles/mixins/list-inline.js
```diff
@@ -0,0 +1,16 @@
+module.exports = function () {
+  return function (mixin) {
+    return {
+      'margin-left': '-5px',
```
jgallen23: what is this for?

styles/mixins/list-inline.js
```diff
@@ -0,0 +1,16 @@
...
+      'padding-left': 0,
+      'list-style': 'none',
+
+      '& > li': {
+        'display': 'inline-block',
+        'padding-right': '5px',
```
jgallen23: we should make this an option

styles/mixins/list-unstyled.js
```diff
@@ -0,0 +1,9 @@
+module.exports = function () {
+  return function (mixin) {
+    return {
+      'padding-left': 0,
```
jgallen23: doesn't need a margin change?

chrisopedia: Why would it?  If you want to reset all of the lists then we could add it, but then you'd have consecutive lists butted up against one another and we have helper classes for spacing to remove margins.  Keeps things separated.

styles/mixins/list-inline.js
chrisopedia: Offsets the padding-left on the `<li />` so you have equal spacing

PR #18, opened 2016-07-19
Create; ugc styles

chrisopedia:

1. provide some default styles for UGC/uncontrolled content

- add in examples to showcase the various styles
- reset some basic styles to match default typographic styles

### DIFF
--- a/examples/ugc/index.html
+++ b/examples/ugc/index.html
@@ -1,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>UGC</title>
+    <link rel="stylesheet" href="../.dist/ugc.css"/>
+  </head>
+  <body>
+    <div class="ugc">
+      <h1>Heading 1</h1>
+      <h2>Heading 2</h2>
+      <h3>Heading 3</h3>
+      <h4>Heading 4</h4>
+      <h5>Heading 5</h5>
+      <p>Default body font</p>
+      <ul>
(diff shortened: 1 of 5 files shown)


styles/helpers/ugc.css
```diff
@@ -0,0 +1,40 @@
+.ugc h1,
```
jgallen23: why not nest?

examples/ugc/index.html
```diff
@@ -0,0 +1,41 @@
+<!DOCTYPE html>
+<html lang="en">
+  <head>
+    <meta charset="utf-8" />
+    <title>UGC</title>
+    <link rel="stylesheet" href="../.dist/ugc.css"/>
```
jgallen23: we might want to test against all of clientkit cause it loads the reset too

styles/helpers/ugc.css
chrisopedia: Didn't seem to be the norm in the rest of the files so I followed the convention.

examples/ugc/index.html
chrisopedia: K.  Good idea.

styles/helpers/ugc.css
jgallen23: lets nest!

styles/helpers/ugc.css
```diff
@@ -0,0 +1,42 @@
+.ugc {
+  & h1,
```
jgallen23: you don't need &

PR #19, opened 2016-07-19
Custom media

jgallen23:

1. to make @media(--tablet) work

Notify or mention any users: @dawnerd

### DIFF
--- a/tasks/css.js
+++ b/tasks/css.js
@@ -25,6 +25,7 @@ const log = new Logr({
 module.exports = function (config, base, outputName, input) {
   const start = new Date().getTime();
   const cssVars = {};
+  const customMedia = {};
 
   Object.keys(config.color).forEach(color => {
     cssVars[`color-${color}`] = config.color[color];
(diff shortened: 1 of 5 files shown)


tasks/css.js
```diff
@@ -36,6 +37,8 @@ module.exports = function (config, base, outputName, input) {
 
   Object.keys(config.breakpoints).forEach(breakpoint => {
     cssVars[`breakpoint-${breakpoint}`] = config.breakpoints[breakpoint]['min-width'];
+    const minmax = (config.breakpoints[breakpoint].smallest) ? 'max' : 'min';
+    customMedia[breakpoint] = `(${minmax}-width: ${config.breakpoints[breakpoint]['min-width']})`;
```
dawnerd: With `max-width` you'd probably want `config.breakpoints[breakpoint]['min-width']` to change. In the new grid I have that bit. Also changed `smallest` to `default` since it wouldn't make since to have desktop be smallest.

jgallen23: yeah, was planning on updating when grid stuff came through

dawnerd: Cool, looks good then!

PR #20, opened 2016-07-20
updates to tabs mixin

jgallen23:

- allow for custom parent class
- es6
- linting
- updated example to show how to style

Notify or mention any users: @chrisopedia

### DIFF
--- a/examples/tabs/tabs.css
+++ b/examples/tabs/tabs.css
@@ -1,1 +1,12 @@
-@mixin tabs 5, 6px;
+.tabs {
+  @mixin tabs 5, 6px;
+
+  & .tab:checked + label {
+    background: #333;
+    color: #fff;
+  }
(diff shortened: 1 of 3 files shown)


jgallen23:
question: should we pass in colors to handle backgrounds and hover styles or leave that to the site?

chrisopedia:
Do we handle colors on clientkit? I don't recall us dealing with colors in mixins. If that's the case I say leave it to the clients site

jgallen23:
Yes. See colors.yaml

chrisopedia:
Yeah. I guess we do have colors in the buttons.

chrisopedia:
Maybe colors we need is just for a border color? Do we have any visuals for tabs we can base it off

chrisopedia:
👍

PR #21, opened 2016-07-20
Update grid to be mobile or desktop first

dawnerd:
Option to be mobile first. Defaults to desktop.

There is one thing Im banging my head against the wall with. Classes like col-md-* have to behave slightly differently depending on which type is chosen. I have it working with desktop first now but Im not sure it'll work the other way around without a lot of code. Would love some ideas.

Also added an example page


jgallen23:
why do md-* need to be different?

conf/default-core.yaml
```diff
@@ -3,3 +3,4 @@ core:
   minimize: false
   dist: '{{CWD}}/.dist'
   assetPath: '{{CWD}}'
+  mobileFirst: false
```
jgallen23: whats the difference between mobileFirst and default: true?

styles/helpers/grid.css
```diff
@@ -73,6 +69,7 @@
   min-height: 1px;
   padding-left: 15px;
   padding-right: 15px;
+  float: left;
```
jgallen23: this seems risky to float all rows. whats the need for it?

conf/default-core.yaml
dawnerd: default: true determines whats the base breakpoint so we're not hardcoding it. mobileFirst handles changing the min/max-width

chrisopedia:
I'm not sure I understand the issue

conf/default-core.yaml
jgallen23: would you ever change one and not the other?

dawnerd:
Before I made the change all of the col-md, col-sm, col-lg etc were based on min width and were mobile first. Now that it can be reversed, col-md doesn't work the same since it falls outside of the breakpoint.

chrisopedia:
The only thing that should change in this scenario is the min width to max and the solves it. The media query shouldn't be invoked on a browser size outside the min width we call desktop

jgallen23:
md==tablet?

conf/default-core.yaml
dawnerd: How would it know which is the default breakpoint? We'd have to sort all the breakpoints by size and compare vs just checking the flag.

jgallen23: good enough reasoning for me. thanks for clarifying

dawnerd: I guess we could also just rename the key to be default like you did on typography

conf/default-breakpoints.yaml
```diff
@@ -1,18 +1,23 @@
 breakpoints:
   desktop-wide:
+    max-width: 100vw
     min-width: 1440px
     content: 1440px
     col: xl
+    default: true
```
jgallen23: what about `{{!mobileFirst}}` and then doing `{{mobileFirst}}` on mobile breakpoint?

dawnerd: That'd work!

PR #23, opened 2016-07-20
Support local mixins

dawnerd:


tasks/css.js
```diff
@@ -47,6 +48,15 @@ module.exports = function (config, base, outputName, input) {
     resolve: m => m(config, postcss)
   });
 
+  if (pathExists.sync(path.join(config.core.assetPath, 'mixins'))) {
```
jgallen23: what about `fs.existsSync`?

jgallen23: nevermind

PR #26, opened 2016-07-21
bg-image and site-max mixins.  removed bleeding

jgallen23:

1. confusion around container-bleed

- remove container-bleed and replaced with bg-bleed
- added example

Notify or mention any users: @dawnerd @ecwillis @chrisopedia @mturnwall

### DIFF
--- a/styles/helpers/bg-bleed.css
+++ b/styles/helpers/bg-bleed.css
@@ -1,9 +1,0 @@
-.bg-bleed {
-  margin-left: auto;
-  margin-right: auto;
-  max-width: 1600px;
-  background-size: cover;
-  background-color: #000;
-  background-repeat: no-repeat;
-  background-position: center center;
-}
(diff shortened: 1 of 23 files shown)


styles/helpers/bg-bleed.css
```diff
@@ -0,0 +1,9 @@
+.bg-bleed {
```
chrisopedia: I've suggested this before but the bleed term seems confusing, can we just change this to `bg-max` or `container-max`, something with max seems more preferred since thats exactly what its doing.

styles/helpers/bg-bleed.css
```diff
@@ -0,0 +1,9 @@
...
+  margin-left: auto;
+  margin-right: auto;
+  max-width: 1600px;
+  background-size: cover;
```
chrisopedia: I feel like this shouldn't be added or assumed because its not always a background image.  We should pass this in maybe as a part of a mixin.

chrisopedia:
👍

PR #30, opened 2016-07-22
print css mixins to console

orthagonal:
just does expressions of the form '@mixin <name>' for right now

### DIFF
--- a/app.js
+++ b/app.js
@@ -5,7 +5,7 @@ const yargs = require('yargs');
 const Logr = require('logr');
 const watcher = require('./lib/watcher');
 const mkdirp = require('mkdirp');
-
+const getStdin = require('get-stdin');
 const log = new Logr({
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -88,15 +83,45 @@ const runAll = () => {
...
     path.join(defaultConf, '*'),
     path.join(argv.config, '*'),
   ];
   watcher(watchedConfigFiles, [''], () => {
     runAll();
   }, 100);
+};
+
+const evalCss = (cssExpression) => {
+  // todo: probably need to actually eval the css expression
```
jgallen23: the @mixin was just an example, I want to eval css.  We probably should read css from stdin.  like `echo some.css | clientkit --css`

jgallen23:
closing. please make separate PRs for css and eslint

PR #32, opened 2016-07-22
Eslints

orthagonal:


conf/default-scripts.yaml
```diff
@@ -0,0 +1,2 @@
+scripts:
+  main.js: '{{CKDIR}}/examples/scripts/sample1.js'
```
jgallen23: please don't add to default scripts.  this should be in examples/clientkit/

tasks/script.js
```diff
@@ -28,6 +28,11 @@ module.exports = function(conf, base, outputName, input) {
     const duration = (end - start) / 1000;
     log(`Processed: ${input} → ${output} in ${duration} sec`);
   });
+  const cli = new CLIEngine({
+    useEslintrc: false,
+    configFile: conf.core.eslint
+  });
+  log(formatter(cli.executeOnFiles([input]).results));
```
jgallen23: if there are no errors are there still results?

jgallen23: would be great to add the error tag to the log function only if there were errors

examples/conf/default-scripts.yaml
```diff
@@ -0,0 +1,3 @@
+scripts:
+  main.js: '{{CKDIR}}/examples/scripts/sample1.js'
```
jgallen23: could use assetsPath here

examples/conf/default-scripts.yaml
```diff
@@ -0,0 +1,3 @@
+scripts:
```
jgallen23: this should be in clientkit directory

PR #39, opened 2016-07-26
Create; initial version of tooltip

chrisopedia:

1. create a CSS only tooltip

- add in examples to showcase
- mixin to generate css for tooltip
- helper to use mixin

image


styles/helpers/tooltip.css
```diff
@@ -0,0 +1,137 @@
+button.tooltip {
```
jgallen23: we shouldn't tie this to button elements

styles/helpers/tooltip.css
```diff
@@ -0,0 +1,137 @@
...
+  overflow: visible;
+}
+.tooltip {
+  @mixin tooltip;
```
jgallen23: we should have the mixin handle all the css.  We need to think of mixins as standalone functions that can work by themselves and then the helper classes just use the mixins

styles/helpers/tooltip.css
```diff
@@ -0,0 +1,137 @@
...
+    &:after {
+      white-space: normal;
+    }
+  }
+
+  &-up {
+    &::before {
+      triangle: equilateral pointing-down;
+      height: 6px;
+      background-color: rgba(17,17,17,.9);
```
jgallen23: we should never hard code colors

styles/helpers/tooltip.css
chrisopedia: its not.  its a hack for ie11

styles/mixins/tooltip.js
```diff
@@ -0,0 +1,35 @@
+module.exports = function () {
+  return function (mixin, color) {
+    var color = color || 'rgba(17,17,17,.9)';
```
jgallen23: can we read this from config? probably want both background and foreground

styles/mixins/tooltip.js
```diff
@@ -0,0 +1,35 @@
...
+        'position': 'absolute',
+        'z-index': 10,
+        'opacity': 0,
+        'border-radius': '4px',
+        'padding': '.5em 1em',
+        'background-color': color,
+        'color': '#fff',
+        'font-size': '12px',
+        'white-space': 'nowrap',
+        'content': 'attr(data-tooltip-title)',
```
jgallen23: this work everywhere?

styles/mixins/tooltip.js
jgallen23: any chance we can get direction passed in here? if no direction passed, then loop through and set everything. That way we can be consistent with our other mixins

jgallen23:
can you add to examples/index.html?

styles/mixins/tooltip.js
chrisopedia: https://developer.mozilla.org/en-US/docs/Web/CSS/attr#Browser_compatibility

- chrome 2.0
- firefox 1.0
- IE 8
- Safari 3.1

styles/mixins/tooltip.js
chrisopedia: The only issue was there was a lot of complex nested and duplicate selectors and I wasn't sure how to do the mixins like that.  For example:

https://github.com/firstandthird/clientkit/pull/39/files#diff-cb4b362339a1552ea99c0267018a7fd6R126

jgallen23: you just return a JavaScript object. it can be nested

jgallen23:
still need to add to examples/index.html and merge in latest from master

jgallen23:
ping

chrisopedia:
Updated to include master but that introduced build errors.  Also added in the examples.

jgallen23:
can't merge

chrisopedia:
@jgallen23 done

jgallen23:
lets update this to the new standard

chrisopedia:
@jgallen23 done

PR #40, opened 2016-07-26
Bg colors

Belelros:
This adds a common way to put utility classes for colours we have as hinted here:

### DIFF
--- a/styles/helpers/utils.css
+++ b/styles/helpers/utils.css
@@ -27,3 +27,4 @@
 }
 
 @mixin hide;
+@mixin bg-colors;
--- a/styles/helpers/utils.css
+++ b/styles/helpers/utils.css
@@ -27,4 +27,3 @@
 }
 
 @mixin hide;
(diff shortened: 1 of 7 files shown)


styles/mixins/bg-colors.js
```diff
@@ -0,0 +1,19 @@
+'use strict';
+
+module.exports = function (config) {
+  return function (mixin) {
```
jgallen23: I like the idea, but I think we can be a little bit more explicit.  In the color object, lets add `background-X` and then have `.bg-X` auto add that background.  If we want to share colors just use something like:

```color:
  background-body: '{{color.someOtherVar}}'
```

styles/helpers/utils.css
```diff
@@ -27,3 +27,4 @@
 }
 
 @mixin hide;
+@mixin bg-colors;
```
jgallen23: lets move this into a `helpers/colors.css`

PR #41, opened 2016-07-26
refactor button

jgallen23:

1. make button mixin more reusable
2. single class for applying button styles
3. naming consistency (more verbose btn vs button)

- change `.btn .btn-primary` to `.button`
- change `.btn .btn-primary-outline` to `.button-outline`
- removed secondary button (projects can add if they need)

Demo: https://firstandthird.github.io/clientkit/buttons/

Notify or mention any users: @mturnwall @dawnerd @chrisopedia

### DIFF
--- a/examples/buttons/index.html
+++ b/examples/buttons/index.html
@@ -20,7 +20,6 @@
       <h1>Buttons</h1>
       <h2 class="heading-3">Default Buttons</h2>
       <button class="button" type="button">Primary</button>
-      <button class="btn btn-secondary" type="button">Secondary</button>
       <h2 class="heading-3">Outline Buttons</h2>
       <button class="button-outline" type="button">Primary</button>
     </section>
--- a/examples/buttons/index.html
+++ b/examples/buttons/index.html
@@ -19,10 +19,10 @@
     <section class="buttons">
       <h1>Buttons</h1>
       <h2 class="heading-3">Default Buttons</h2>
-      <button class="btn btn-primary" type="button">Primary</button>
+      <button class="button" type="button">Primary</button>
       <button class="btn btn-secondary" type="button">Secondary</button>
       <h2 class="heading-3">Outline Buttons</h2>
-      <button class="btn btn-primary-outline" type="button">Primary</button>
+      <button class="button-outline" type="button">Primary</button>
     </section>
   </main>
   <footer>
--- a/styles/mixins/button.js
+++ b/styles/mixins/button.js
@@ -1,15 +1,42 @@
+'use strict';
 module.exports = function () {
-  return function (mixin, bgColor, color) {
-    const activeBgColor = 'red';
-    const styles = {};
-    const state = {
-      'background-color': activeBgColor
-    };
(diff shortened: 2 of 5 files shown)


styles/mixins/button.js
```diff
@@ -1,15 +1,42 @@
...
-    };
-    styles['background-color'] = bgColor;
-    styles.color = color;
+  return function (mixin, bgColor, fgColor, type) {
+
+    const styles = {
+      display: 'inline-block',
+      'line-height': 1,
+      'text-align': 'center',
+      'font-size': '14px',
```
chrisopedia: shouldn't this be passed in from the config?  Client might have a larger button and the font-size will help determine that.

styles/mixins/button.js
```diff
@@ -1,15 +1,42 @@
...
+      'vertical-align': 'middle',
+      cursor: 'pointer',
+      transition: 'color 250ms cubic-bezier(0.250, 0.460, 0.450, 0.940), background-color 250ms cubic-bezier(0.250, 0.460, 0.450, 0.940)',
+      color: fgColor,
+    }
+
+    let state = {};
+    if (type === 'outline') {
+      styles['background-color'] = 'transparent';
+      styles.border = `2px solid ${bgColor}`;
```
chrisopedia: Should we put the thickness of the border as an option?

examples/buttons/index.html
```diff
@@ -19,10 +19,10 @@ <h1 class="heading-1">ClientKit Demos</h1>
     <section class="buttons">
       <h1>Buttons</h1>
       <h2 class="heading-3">Default Buttons</h2>
-      <button class="btn btn-primary" type="button">Primary</button>
+      <button class="button" type="button">Primary</button>
       <button class="btn btn-secondary" type="button">Secondary</button>
       <h2 class="heading-3">Outline Buttons</h2>
-      <button class="btn btn-primary-outline" type="button">Primary</button>
+      <button class="button-outline" type="button">Primary</button>
```
chrisopedia: My only concern with having no base style for a button is that you're duplicating a lot of the common code that exists for buttons.

styles/mixins/button.js
```diff
@@ -1,15 +1,42 @@
...
-    styles['background-color'] = bgColor;
-    styles.color = color;
+  return function (mixin, bgColor, fgColor, type) {
+
+    const styles = {
+      display: 'inline-block',
+      'line-height': 1,
+      'text-align': 'center',
+      'font-size': '14px',
+      padding: '12px 14px',
```
chrisopedia: Are we gonna have different sizes for buttons?  If so we should pass in the options for padding to increase the size.

examples/buttons/index.html
jgallen23: I guess.... just seems annoying to have to put two classes for a button

styles/mixins/button.js
jgallen23: as of right now, we only have one size.  if we add another, we could add another param to the mixin

dawnerd:
👍

PR #45, opened 2016-07-28
Tests

orthagonal:


test/test.mixins.js
```diff
@@ -0,0 +1,201 @@
...
+const path = require('path');
+
+// global config object for each test
+const conf = require('confi')({
+  path: [
+    './conf'
+  ],
+  context: {
+    CKDIR: __dirname,
+    CONFIGDIR: path.join(process.cwd(), 'clientkit')
```
jgallen23: is this right?

test/test.mixins.js
```diff
@@ -0,0 +1,201 @@
...
+  });
+});
+
+describe('buttons mixin', function() {
+  this.timeout(20000);
+  it('generates css for buttons', (done) => {
+    const button = require('../styles/mixins/button.js')(conf);
+    // mixin, bgColor, fgColor, type
+    const result = button({}, '#ff0000', '#00ff00', 'outline');
+    expect(result.color).to.equal('#ff0000');
```
jgallen23: can we do a full `expect(result, { ... })`

test/test.mixins.js
```diff
@@ -0,0 +1,201 @@
...
+    expect(result.border).to.equal('2px solid #ff0000');
+    expect(result['&:hover'].color).to.equal('#00ff00');
+    done();
+  });
+});
+describe('buttons outline mixin', function() {
+  this.timeout(20000);
+  it('generates css outlines for buttons', (done) => {
+    const button = require('../styles/mixins/button-outline.js')(conf);
```
jgallen23: make sure you are looking at master, I don't think button-outline exists anymore

jgallen23: oh, I guess it does, we should probably delete

test/test.mixins.js
jgallen23: what I think we should do is have an output directory with the expected output of each mixin as json and then just do a comparison.  would make it easier to handle updates on the tests

test/test.mixins.js
```diff
@@ -0,0 +1,181 @@
...
+  it('generates css bg image', (done) => {
+    const image = require('../styles/mixins/bg-image.js')(conf);
+    const result = image({}, 'http://nope.com/myImage.jpg');
+    compare(result, 'bg-image.js');
+    done();
+  });
+});
+
+describe('buttons mixin', function() {
+  this.timeout(20000);
```
jgallen23: whoa, takes a long time to run the tests?

test/test.mixins.js
```diff
@@ -0,0 +1,181 @@
...
+  it('generates css for buttons', (done) => {
```
jgallen23: we should have tests for non-outline

test/test.mixins.js
```diff
@@ -0,0 +1,181 @@
...
+  it('generates grid classes', (done) => {
+    const grid = require('../styles/mixins/grid.js')(conf);
+    const result = grid({});
+    compare(result, 'grid.js');
+    done();
+  });
+});
+
+describe('hide', function() {
+  it('generates hide classes', (done) => {
```
jgallen23: can we do the compare here?

test/test.mixins.js
```diff
@@ -0,0 +1,181 @@
...
+  });
+});
+describe('spacing mixin', function() {
+  this.timeout(5000);
+  it('generates basic css spacers', (done) => {
+    const spacing = require('../styles/mixins/spacing.js')(conf);
+    const result = spacing({}, 'margin', 'top', 'md');
+    expect(result['margin-top']).to.equal(conf.spacing.default.md);
+    const result2 = spacing({}, 'padding', 'bottom', 'none');
+    expect(result2['padding-bottom']).to.equal('0');
```
jgallen23: can you check the entire result?  wouldn't want other stuff to be in there

test/test.mixins.js
```diff
@@ -0,0 +1,181 @@
...
+  //   // expect(result.css).to.include('margin-left');
+  //   // expect(result.css).to.include('margin-right');
+  //   return done();
+  // });
+});
+*/
+it('generates basic css tabs', (done) => {
+  const tabs = require('../styles/mixins/tabs.js')(conf);
+  const result = tabs({}, '3', '12');
+  expect(result['border-radius']).to.equal('12');
```
jgallen23: can we do compare here?

PR #47, opened 2016-07-28
fixed spacing issue

orthagonal:


styles/mixins/spacing.js
```diff
@@ -14,8 +14,22 @@ module.exports = function (config) {
     return breakpointHelper(styles, config);
   };
 
-
+  const axials = {
```
jgallen23: I know this goes against what I originally said, but I think this logic belongs in the function above, so you do the position check there. would that make things cleaner?

orthagonal: Yeah I think it makes more sense that way, the outer mixin call is really just attaching a css node, it shouldn't care how many children that node has.

jgallen23:
this is so much cleaner!  I should have just listened to you in the first place 😛

PR #48, opened 2016-07-28
Create; mixin and helper for text-stretch

chrisopedia:

1. common pattern seen across designs

- create mixin
- create helper
- update example


styles/helpers/typography.css
```diff
@@ -30,6 +30,10 @@ body {
   text-align: right;
 }
 
+.text-stretch {
+  @mixin text-stretch 2px;
```
jgallen23: how consistent is 2px all the time?  should this be project specifc?

chrisopedia: Could be. So far from the few designs I've seen it looks like this is the norm.

jgallen23: feels like the 2px should be a var driven by the config and then sites could override the yaml var if they wanted to tweak it

jgallen23:
if you are going to do text stretch, just add this to fontStyles:

`text-stretch: 
  letter-spacing: 2px
`
and that'll auto create a helper class.  Make sure you are updating the tests also.  run `npm test`

styles/helpers/typography.css
```diff
@@ -30,6 +30,10 @@ body {
   text-align: right;
 }
 
+.text-stretch {
```
jgallen23: don't need this anymore

styles/mixins/text-stretch.js
```diff
@@ -0,0 +1,8 @@
+module.exports = function ( config ) {
```
jgallen23: you can kill this

test/expectedOutputs/mixins/font-style.js
```diff
@@ -78,5 +78,11 @@ module.exports={
       "font-size": "12px",
       "line-height": "22px"
     }
+  },
+  ".text-stretch": {
+    "letter-spacing": "2px",
+    "@media (max-width: 767px)": {
```
jgallen23: hmm, we should put a check in here so it doesn't output if it's not there

jgallen23: I'll add a ticket for it

PR #53, opened 2016-07-29
adds '--minify' or 'minify' CLI option

orthagonal:

1.
2.
3.

-
-
-

Notify or mention any users:

Fixes: #<issue number>

-added option and support for it
-added test for minify
-corrected 'minimize' -> 'minify' in default-core.yaml
-includes correction to esnure it passes travis test

### DIFF
--- a/app.js
+++ b/app.js
@@ -16,10 +16,6 @@ const log = new Logr({
 });
 
 const argv = yargs
-.option('minify', {
-  describe: 'minify the css/js output. useful when generating production code',
-  default: false
-})
 .option('options', {
   describe: 'shows the css variables and mixins that are available ',
   default: false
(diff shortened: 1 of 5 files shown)


app.js
```diff
@@ -59,7 +64,10 @@ const loadConfig = () => {
       CONFIGDIR: argv.config
     }
   });
-  // will need mode later on:
+  // second, set up the configuration based on any command line options:
+  if (argv.minify || argv._.minify || argv._.indexOf('minify') > -1) {
+    conf.minify = true;
```
jgallen23: how bout just setting conf.core.minify so we don't have two variables to check

tasks/css.js
```diff
@@ -104,7 +104,8 @@ class CssTask {
         foundries: ['custom', 'hosted', 'google']
       }));
     }
-    if (this.config.core.minify || this.config.mode === 'prod') {
+    // minify if specified in config files:
+    if (this.config.core.minify) {
```
jgallen23: missing the check on scripts

PR #55, opened 2016-07-30
verifies config dir exists, exits with error if not

orthagonal:


app.js
```diff
@@ -49,6 +50,13 @@ const cssProcessor = require('./tasks/css.js');
 const jsProcessor = require('./tasks/script.js');
 
 const loadConfig = () => {
+  try {
+    fs.accessSync(argv.config, fs.F_OK);
+  } catch (e) {
+    log(['error'], `Unable to access config directory ${argv.config}`);
+    log(e);
```
jgallen23: Don't need both logs

app.js
```diff
@@ -53,8 +53,7 @@ const loadConfig = () => {
   try {
     fs.accessSync(argv.config, fs.F_OK);
   } catch (e) {
-    log(['error'], `Unable to access config directory ${argv.config}`);
-    log(e);
+    log(['error'], e);
```
jgallen23: Just confirming this will log the actual error. I've had issues stringifying errors

jgallen23: http://stackoverflow.com/questions/34298063/why-does-json-stringify-on-typeerror-return-an-empty-object

jgallen23:
Need to add tests

PR #57, opened 2016-08-02
will render an html styleguide

orthagonal:


lib/styleguide.template.js
```diff
@@ -0,0 +1,103 @@
+'use strict';
+const _ = require('lodash');
```
jgallen23: FYI. all the lodash functions you used are built into node

lib/styleguide.template.js
```diff
@@ -0,0 +1,103 @@
...
+
+module.exports = (cssVars) => {
+return `
+<html>
+  <head> </head>
+  <body>
+    <h1> Styleguide </h1>
+    <h2> Colors <h2>
+    <table>
+      ${_.reduce(cssVars, (memo, value, name) => {
```
jgallen23: would a map work better?

jgallen23: nm, you have the if statement

tasks/css.js
```diff
@@ -2,7 +2,7 @@
 
 const fs = require('fs');
 const path = require('path');
-
+const _ = require('lodash');
```
jgallen23: you added lodash, but aren't using it.  make sure to lint!

tasks/css.js
```diff
@@ -139,12 +139,27 @@ class CssTask {
     log(`Wrote: ${this.input} → ${output}`);
     log(`Wrote: ${this.input}.map → ${output}.map`);
   }
+
+  exportStyleguide(inputName, outputName) {
+    try {
+      const htmlTemplate = require(path.join(process.cwd(), 'lib', 'styleguide.template.js'));
```
jgallen23: this should be __dirname, not cwd()

tasks/css.js
```diff
@@ -139,12 +139,27 @@ class CssTask {
...
+      const output = htmlTemplate(this.cssVars);
+      fs.writeFileSync(outputName, output);
+    } catch (e) {
+        console.log(e)
```
jgallen23: `log(['error']...`

orthagonal: logr 1.10.0 will add an 'error' tag when you pass it an Error object  :-)

PR #58, opened 2016-08-05
Flex grid

Belelros:

1. Provides a flexbox grid option for Clientkit.


jgallen23:
should we use the same class names as the normal grid?  that way we can just drop in the fallback for non-flexbox browsers?

can you also make sure tests are added?

Belelros:
I'll add the tests. Regarding class names, I thought about that and, after that I discarded the idea. Having flex frees you from the need of having to add a row per line and sometimes or we'll use it for column layout and non-flex grid won't supply anything that mimics that. 

That said, flexbox is supported from IE10 and upwards (with some bugs but nothing impossible). 

Do you still think there's a need for that?

jgallen23:
k, lets skip the classname matching for now.

styles/default.css
```diff
@@ -3,6 +3,7 @@
 @import 'helpers/box-sizing.css';
 @import 'helpers/colors.css';
 @import 'helpers/grid.css';
+@import 'helpers/flex-grid.css';
```
jgallen23: looks like this was added twice

styles/helpers/flex-grid.css
```diff
@@ -0,0 +1,5 @@
+.flex {
+  flex: 1;
+}
+
+@mixin flex-grid;
```
jgallen23: should we include this by default?

styles/helpers/flex-grid.css
```diff
@@ -0,0 +1,5 @@
+.flex {
```
jgallen23: should this be in the mixin?

Belelros: Since this wasn't dependant on size nor layout I thought that's the only thing that made sense to leave out of the mixin. However we can add it for consistency's sake.

styles/helpers/flex-grid.css
Belelros: What do you mean by default?

jgallen23: just wondering if it should be an opt in for flex box or included in clientkit by default

Belelros: I guess that would make sense nowadays.

chrisopedia:
# Feedback

The major issue I'm seeing is the responsiveness, but the rest of the stuff listed below is just a personal preference and my initial gut response to the examples page.

## Responsive

Looking at the example and somethings not adding up on the responsive side.  Using chrome dev tools at the 768px I'm getting horizontal scroll bars; this continues all the way down to the mobile sizes as well.  I tried debugging, but I'm not coming up with whats going on to cause this.

## layout vs row

We're using the `.layout` class as essentially the equivalent of `.row` on the fluid-grid so why not use the same name or if that causes conflict, just call it `flex-row` for consistency sake.  I think this will provide a consistency in the naming as well as allowing someone to know that this class is an extension of the flexbox rather than something that can be independently added.

## start, end and center

I feel that using `start` and `end` as the modifiers might be a bit misleading for a newcomer to flex (even though it matches the properties) when all you're doing is left aligning and right aligning respectively.  Might be better if we named it `flex-align-left` instead of the verbose and confusing `layout-start-start`.  It also coincides better with the `center` modifier.  Then you'd have `*-left`, `*-center` and `*-right`.

## order numbering

Wondering if there is a better or more clear way to name the ordering of things.  Given with flexbox you can only really change the first and last items, it might make more sense to name it `flex-order-first` and `flex-order-last` rather than the `flex-order--1` and the `flex-order---1`.  With the latter syntax I feel you're more likely to make an error because you're unsure of how many dashes are required to get the desired effect.

Belelros:
Great feedback @chrisopedia will address it shortly.

PR #61, opened 2016-08-16
Styles for text based inputs

dawnerd:
Thoughts on the text style inputs?

Also, how do we want to handle the other cases like select and checkboxes?

Screenshot attached. Not pictured is active state.
image


jgallen23:
this is a good start. going to merge so I can start using, but we should  fix tests

PR #62, opened 2016-08-17
added icon mixin

jgallen23:

1. gives us a stock set of icons to quick add to projects

- added `@mixin icon`
- adds ionicons to the repo

Notify or mention any users:

@dawnerd @mturnwall @chrisopedia @ecwillis @Belelros


jgallen23:
Diff is large so will summarize:

Usage: 

screenshot 2016-08-17 at 10 36 07 am

Output:

screenshot 2016-08-17 at 10 36 39 am

Icons available: http://ionicons.com/

chrisopedia:
1. Are we sure that ion will be able to provide all the necessary icons we'll need?  Are they still adding to the list?  Looking at the repo doesn't show a lot of activity and there are a lot of requests for new icons that have gone unnoticed.
2. What does `svg-load()` output as its final CSS?  I noticed its a background image and I think it encodes it as Base64, but does it add in all of the other attributes to prevent it from repeating?
3. Base64 has [some performance considerations](http://dev.mobify.com/blog/data-uris-are-slow-on-mobile/) for mobile users
4. You might want to add the `vertical-align: middle` to your CSS mixin to align it with text since that's mostly whats going to be used.

jgallen23:
1) we can easily add other libraries.  I just find ionicons look the best.  there are 700+, so I think we're good for now
2) Using svg-load isn't new to this PR.  As far as repeating, the image is set to the same size as the container, so no issues there
3) did you see the update? http://dev.mobify.com/blog/base64-does-not-impact-data-uri-performance/
4) k, I'll add

mturnwall:
I think this implementation is clean and easy to use.

I'm not sure how ionicons does alignment but a service like Icomoon aligns the top of the icon to the ascender line. This means there is no need for `vertical-align` because the height of the icon is now determined by the height of the font based on em.

https://icomoon.io/#docs/font-metrics
> The IcoMoon app aligns the top of the icon canvas to the ascender line and its bottom to the descender line of the font. Using this technique, the size of your font will directly translate to the size of the icon. For example, if you set your font-size in CSS to be 16px, you will get a 16px icon. This method was first introduced by IcoMoon.

I'd look at Icomoon for the library. They have a ton of free icons and some other packs for cheap.

Also is there a way to only add the icons needed?

jgallen23:
is that for icon fonts or svgs?

jgallen23:
just added support for local files as well:

`@mixin icon {path},{size}`

mturnwall:
Icomoon does both icon fonts and svg. It gives you the option of which version you want svg or an icon font.

image

jgallen23:
what does their css look like if you use svgs?  maybe we just steal that

PR #67, opened 2016-08-18
Close

chrisopedia:
1. because greg said so, haha
2. provides a close class for modals or notifications or anything that would require a close button so its standardized

- create a helper that provides a hover and focus state

Notify or mention any users:

@jgallen23


jgallen23:
two things.  
1) can we add inline documentation and remove example?
2) can we set the content of it? or do I need to copy and paste the nice little x every time?

chrisopedia:
1. done and pushed
2. I don't really have a good reason for or against except to say I've never seen anyone do that which makes me feel like having the text inside is required.

jgallen23:
kill the example html

chrisopedia:
done

PR #68, opened 2016-08-18
Col plain

chrisopedia:

1. fixes the issue with `.col-plain` not being specific enough
2. provides the ability to add `.col-plain` on various devices screens

- convert the col-plain helper to part of the mixin to give it equal specificity
- adds in unit tests

Notify or mention any users:

@jgallen23 @Belelros


jgallen23:
looks good to me

Belelros:
Looks good to me too.

PR #70, opened 2016-08-19
Fontstyles Refactor

jgallen23:

1. getting rid of the yaml magic
2. starting to move mixins to css declared

- removing typography.yaml and using yaml for variables only


jgallen23:
screenshot 2016-08-22 at 1 17 22 pm

Belelros:
LGTM. Really welcomed changes, lighter styles. `font-primary` and similar carried colour too.

jgallen23:
for color, should we add a `color-body`? helper class?  I was thinking of doing a `color-light` for use on dark backgrounds.  or more verbose. `font-color-body`?

also, should we had font-weights to headlines?

styles/mixins/headings.css
```diff
@@ -0,0 +1,35 @@
...
+}
+@media (--mobile) {
+  .heading-1 {
+    font-size: 30px;
+  }
+}
+```
+*/
+
+@define-mixin heading $size, $mobileSize {
```
jgallen23: another option I could do is have everything allowed to be passed in and set the global vars as the defaults.

`@define-mixin heading $size, $mobileSize, $family: var(--font-heading), $color: var(--color-heading)...)`

dawnerd:
Anything left to get this merged?

jgallen23:
Wanted to get at least a few eyes on it.  Can probably pull in tonight or
tomorrow

conf/default-vars.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+  #weights
+  font-light: 300
+  font-regular: 400
+  font-semibold: 600
+  font-bold: 700
+  font-xbold: 800
+
+  #headings
+  heading-line-height: 120%
```
chrisopedia: can we convert this to 1.2?

conf/default-vars.yaml
```diff
@@ -0,0 +1,35 @@
+vars:
+  #families
+  font-primary: 'sans-serif'
+  font-secondary: 'serif'
+  font-heading: 'serif'
```
chrisopedia: do we need to add one for monospaced fonts?

conf/default-vars.yaml
```diff
@@ -0,0 +1,35 @@
...
+
+  line-height: 1.1
+  line-height-large: 1.5
+
+  #weights
+  font-light: 300
+  font-regular: 400
+  font-semibold: 600
+  font-bold: 700
+  font-xbold: 800
```
chrisopedia: is this necessary?  seems like we're just listing out all of the weights and not all font-families we'll support these specific numbers

conf/default-vars.yaml
```diff
@@ -0,0 +1,35 @@
...
+  heading-2-size: 30px
+  heading-2-mobile-size: 25px
+  heading-3-size: 25px
+  heading-3-mobile-size: 20px
+  heading-4-size: 16px
+  heading-4-mobile-size: 14px
+  heading-5-size: 14px
+  heading-5-mobile-size: 14px
+
+  stretch-spacing: 2px
```
chrisopedia: might want to change this to `letter-spacing` just so its explicit

styles/helpers/typography.css
```diff
@@ -123,18 +119,22 @@ title: Weight
...
+<p class="font-large">Some Text</p>
+<p class="font-small">Some Text</p>
 ```
 */
-@mixin font-style;
+.font-large {
+  font-size: var(--font-size-large);
+}
+
+.font-small {
```
chrisopedia: should we attach the `<small />` element to this as well?

styles/mixins/headings.css
```diff
@@ -0,0 +1,35 @@
...
+  font-family: var(--font-heading);
+  color: var(--color-heading);
+  line-height: var(--heading-line-height);
+  font-size: $size;
+  @media (--mobile) {
+    font-size: $mobileSize;
```
chrisopedia: won't the `line-height` change here as well?

styles/helpers/ugc.css
```diff
@@ -51,27 +51,23 @@ title: UGC
...
   & h4 {
-    @mixin font-style heading-4;
+    @mixin heading var(--heading-4-size), var(--heading-4-mobile-size);
   }
   & h5 {
-    @mixin font-style heading-5;
-  }
-  & p {
-    @mixin font-style font-body;
+    @mixin heading var(--heading-5-size), var(--heading-5-mobile-size);
```
chrisopedia: this should use the `font-body` var so we're not tied to the h5 in case we want to set those differently.  I'm not even entirely sure we need to set the font-size on the `<p />` at all since it will inherit from the body.

conf/default-vars.yaml
chrisopedia: Should we also just pick the generically available fonts like Arial and create a prettier font stack that can be overridden?  Or do we get super fancy and add the system font declarations?

jgallen23: RE mono.  I think that could be added in a separate PR.

In terms of font family. I think its assumed that the parent site would always override these.  I kinda like the idea of it defaulting to the system font declarations though. maybe that could be a separate PR

conf/default-vars.yaml
jgallen23: was just ported over from the previous and used in the helpers.  I guess we can probably simplify to just 3 sizes

conf/default-vars.yaml
jgallen23: but letter spacing for what?  it's specifically for the stretch class

styles/helpers/typography.css
jgallen23: great question. was wondering if we should be styling elements in here as well

styles/mixins/headings.css
jgallen23: base line height will be 1.2. probably doesn't need to change on mobile

styles/helpers/ugc.css
jgallen23: look at the diff closer, it's applying to h5

conf/default-vars.yaml
chrisopedia: I think I'm more getting at leaving this out completely, or including all of the possibilities.  This just seems a bit misleading or incomplete.

conf/default-vars.yaml
chrisopedia: True.  But `stretch-spacing` feels like it could be used for more than it is, which we may want but probably not.

styles/helpers/typography.css
chrisopedia: I'd say so.  But thats just me.

styles/helpers/ugc.css
chrisopedia: :doh

styles/mixins/headings.css
chrisopedia: Hard to assume that given the line-height can be changed from the client so if they go bigger fonts, than 1.2 will probably not work.  Something that can be revisited, but just wanted to address it.

conf/default-vars.yaml
chrisopedia: k

PR #75, opened 2016-08-22
test whether changes to clientkit affect the default output

orthagonal:


test/test.defaultCss.js
```diff
@@ -0,0 +1,31 @@
+'use strict';
+const describe = require('mocha').describe;
+const it = require('mocha').it;
+const expect = require('chai').expect;
+const path = require('path');
+const fs = require('fs');
+const cssModule = require('../tasks/css.js');
+
+const conf = require('confi')({
```
jgallen23: can we reuse the app.js instead?

test/test.defaultCss.js
```diff
@@ -0,0 +1,31 @@
...
+    CONFIGDIR: path.join(process.cwd(), 'conf')
+  }
+});
+
+describe('default css output', function() {
+  this.timeout(15000);
+  it('detects dev changes that affect the expected CSS output', (done) => {
+    // generate css against ./styles/default.css
+    const cssTask = new cssModule.CssTask(conf, process.cwd());
+    cssTask.performTask('./styles/default.css', (result) => {
```
jgallen23: we should run this with the NODE_ENV=test so we can disable minify so we get better comparison diffs

jgallen23: I also wonder if it would be better to break it apart to compare each individual files instead of the whole thing.  or maybe both?

jgallen23:
ping on this

PR #78, opened 2016-08-23
makes sourcemap generation optional

orthagonal:
'on' will generate a sourcemap file and embed a ref to it at the end of
the output file
'off' will not generate a sourcemap
'inline' will generate a sourcemap and embed it directly at the end of
the output file


app.js
```diff
@@ -38,6 +38,10 @@ const argv = yargs
   describe: 'a path to your configuration files',
   default: path.join(process.cwd(), 'clientkit')
 })
+.option('sourcemap', {
+  describe: 'determines how/whether css/js sourcemaps are generated, can be one of "on", "off" or "inline"',
+  default: 'on'
```
jgallen23: can we default to the option in the conf?

lib/removeSourceMap.js
```diff
@@ -0,0 +1,10 @@
+// removes embedded source maps in output css/js files:
+module.exports = (text, type) => {
```
jgallen23: why do we need this?  if we don't run through sourcemap process or enable debugging, this shouldn't get added, correct?

tasks/script.js
```diff
@@ -61,12 +71,16 @@ module.exports = function(conf, base, outputName, input) {
   if (conf.core.minify) {
     currentTransform = currentTransform.transform(uglifyify, { global: true });
   }
-  currentTransform
+  const stream = currentTransform
```
jgallen23: debug needs to be set to false above if sourcemaps are off.

tasks/script.js
```diff
@@ -61,12 +71,16 @@ module.exports = function(conf, base, outputName, input) {
...
   .bundle()
   .on('error', function (err) {
     log(['error'], err);
     this.emit('end');
-  })
-  .pipe(exorcist(`${output}.map`))
-  .pipe(fileStream);
+  });
+  if (conf.sourcemap === 'on') {
+    stream.pipe(exorcist(`${output}.map`))
```
jgallen23: does inline work for javascript?

app.js
jgallen23: actually, lets just set it in the config yaml instead of cli arg

lib/removeSourceMap.js
orthagonal: I tried to make this happen, but neither postcss and browserify seem to support a basic option to stop them from planting a sourcemap tag at the end of the css/js file.  Just the power to either embed the sourcemap directly or as a link to the external file.

tasks/script.js
orthagonal: Got it.

tasks/script.js
orthagonal: Yup.

PR #80, opened 2016-08-24
Create; helper for table with various properties

chrisopedia:

~~❗ this is not mergeable as it relies on #70~~

1. adds in support for various versions of tables
2. includes documentation in mdcss form

- small version
- bordered variant
- hover treatment
- striped version to give zebra striping
- responsive container

## Screenshot

image


conf/default-vars.yaml
```diff
@@ -33,3 +33,13 @@ vars:
   heading-5-mobile-size: 14px
 
   stretch-spacing: 2px
+
+  table-cell-padding: .75rem
```
jgallen23: seems weird to start using rems now

conf/default-vars.yaml
```diff
@@ -33,3 +33,13 @@ vars:
...
+  table-sm-cell-padding: .3rem
+  table-bg: transparent
+  table-bg-accent: '#f9f9f9'
```
jgallen23: colors should be in colors.yaml

jgallen23: unless we want to start moving colors to be in context with the vars.

@dawnerd thoughts?

jgallen23:
can you post screenshot of docs?

chrisopedia:
changed to pixels and added screenshot

PR #83, opened 2016-08-24
Delete broken tests for old font-styles mixin

chrisopedia:

1. all builds for new features have broken tests

- delete test
- delete mixin test suite

Notify or mention any users: @jgallen23 @dawnerd


dawnerd:
Looks good. We'll probably want to bring the tests back in at some point.

PR #84, opened 2016-08-24
Update; convert row-carousel mixin to use helper

chrisopedia:

1. simplify the mixins to CSS instead of JS

- conver js mixin to css
- use mixin to support responsive

Notify or mention any users: @jgallen23 @dawnerd


styles/mixins/row-carousel.css
```diff
@@ -0,0 +1,19 @@
+@define-mixin row-carousel {
```
jgallen23: shouldn't this be a helper instead?

chrisopedia: what you're not seeing is the helper file didn't change since it was a bunch of responsive classes calling the same mixin.  This prevents us from writing the same code for 4 different classes.

chrisopedia: https://github.com/firstandthird/clientkit/blob/row-carousel-helper/styles/helpers/carousel.css

jgallen23: hot damn. makes sense

PR #89, opened 2016-08-24
added message helpers and mixins

jgallen23:

1. common way to display success/error/info messages on pages

- added `.message-info` `.message-success` `.message-error` helper classes
- added `@mixin message [color];` mixin

Notify or mention any users: @firstandthird/developers


jgallen23:
screenshot 2016-08-24 at 9 29 04 am

conf/default-colors.yaml
```diff
@@ -20,3 +20,6 @@ color:
   button-primary-bg: 'rgba(75, 157, 249, 1)'
   button-primary-color: '{{color.white}}'
   icons: '#333'
+  message-info: '#f0f0f0'
+  message-success: '#dff0d8'
+  message-error: '#f2dede'
```
chrisopedia: might want to change these names to be a bit more generic since they'll be used in text elements, alert boxes, notifications, form validation, etc.

jgallen23: that's why I used message.  that's about as generic as I could come up with.

dawnerd:
Looks good to me.

Re: message vars - input has color-input-error. Might be able to share a var there?

PR #92, opened 2016-08-27
removes 'styleguide' from being generated

orthagonal:


lib/styleguide.template.js
```diff
@@ -1,122 +0,0 @@
-'use strict';
-const reduce = require('lodash.reduce');
```
jgallen23: are these still required in package.json?

orthagonal: We still use it in commands/dev.js

jgallen23: capitalize too?

PR #94, opened 2016-09-01
Flattenvars

orthagonal:


conf/default-vars.yaml
```diff
@@ -1,48 +1,51 @@
...
+    regular: 400
+    semibold: 600
+    bold: 700
+    xbold: 800
+    #sizes
+    size:
+      body: 14px
+      large: 16px
+      small: 12px
+  line:
```
jgallen23: probably don't need to split this one

PR #96, opened 2016-09-02
notifies user of any new updates to clientkit

orthagonal:


app.js
```diff
@@ -48,32 +51,47 @@ const argv = yargs
...
-} else if (argv.mode === 'dev' || argv._.dev || argv._.indexOf('dev') > -1) {
-  dev.runDev(defaultConf, conf, argv, log);
-// normal mode will run and output the new css/js dist directory:
-} else {
-  if (argv.debug || argv._.indexOf('debug') > -1) {
-    log(JSON.stringify(conf, null, '  '));
+const handleNotifier = (err, updateResults) => {
+  if (!err) {
+    if (updateResults.latest !== updateResults.current) {
+      log(`A new version of clientkit is available on npm! Current: ${updateResults.current}  Latest: ${updateResults.latest}`);
```
jgallen23: lets add a warning tag

app.js
```diff
@@ -48,32 +51,47 @@ const argv = yargs
...
+    }
   }
-  run.runAll(conf, log);
-}
+  main();
+};
+
+updateNotifier({
+  callback: handleNotifier,
```
jgallen23: > Passing a callback here will make it check for an update directly and report right away. Not recommended as you won't get the benefits explained in How.

orthagonal: For the most part, I think we want to check for an update directly and report right away and then move on to the main clientkit logic. The only reason to do it the other way is if we want to keep periodically checking when running clientkit in dev mode.  We can totally do it that way, but seemed like overkill to me.

jgallen23: we restart clientkit a lot, so I wouldn't want it to slow down the main process or check all the time.

app.js
```diff
@@ -48,32 +51,42 @@ const argv = yargs
...
+    if (argv.debug || argv._.indexOf('debug') > -1) {
+      log(JSON.stringify(conf, null, '  '));
+    }
+    run.runAll(conf, log);
+  }
+};
+
+main();
+const result = updateNotifier({
+  updateCheckInterval: 1000 * 60 * 60, // check once per hour
```
jgallen23: no need to continue to check

PR #103, opened 2016-09-03
Create input group

chrisopedia:

1. a common pattern we use on footers for newsletter sign ups (also search)

- adds a new input-group-background var for color
- adds in the mdcss
- adds in the section for the helper

Screenshot of docs:

image

### DIFF
--- a/styles/helpers/inputs.css
+++ b/styles/helpers/inputs.css
@@ -178,7 +178,7 @@ title: Form
       margin-right: -1px;
       padding-right: 14px;
       padding-left: 14px;
-      border-radius: var(--border-radius) 0 0 var(--border-radius);
+      border-radius: var(--border-radius-default) 0 0 var(--border-radius-default);
       background-color: var(--color-input-group-background);
       font-size: var(--font-size-body);
 
@@ -189,7 +189,7 @@ title: Form
     & .input-group-btn {
       display: inline-block;
       vertical-align: middle;
-      border-radius: 0 var(--border-radius) var(--border-radius) 0;
+      border-radius: 0 var(--border-radius-default) var(--border-radius-default) 0;
     }
   }
 }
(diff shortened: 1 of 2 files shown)


styles/helpers/inputs.css
```diff
@@ -140,4 +140,56 @@ title: Form
...
+    & input[type="time"] {
+      display: inline-block;
+      width: inherit;
+      height: 42px;
+      vertical-align: middle;
+      box-shadow: none;
+      margin-right: -1px;
+      padding-right: 14px;
+      padding-left: 14px;
+      border-radius: var(--border-radius) 0 0 var(--border-radius);
```
jgallen23: I think we need to update this based on the other PR

styles/helpers/inputs.css
```diff
@@ -140,4 +140,56 @@ title: Form
...
+
+  /*---
+  section: Forms
+  title: Input group
+
+  ```example:html
+  <form class="form">
+    <div class="input-group">
+      <input type="text" placeholder="Enter your email">
+      <button class="input-group-btn button">Sign Up</button>
```
jgallen23: do you need the `input-group-btn` class?  Can you just assume there is a .button in there?

chrisopedia: Yes, because we're resetting the border-radius on the button and we want that regardless of whether we use `.button` or not.

styles/helpers/inputs.css
chrisopedia: what PR?

styles/helpers/inputs.css
jgallen23: what else would we use?

chrisopedia: spans for addons like currency or things you'd wnat to suffix with.  See [Bootstrap](http://v4-alpha.getbootstrap.com/components/input-group/#basic-example)

styles/helpers/inputs.css
jgallen23: The one where you changed this variable

styles/helpers/inputs.css
jgallen23: Then should we not call it button?

chrisopedia: I think the way it is, provides more flexibility.

PR #104, opened 2016-09-03
Create dropdown helper

chrisopedia:

1. abstract out the common dropdown pattern into something generic

- new variable for the background
- create the mdcss docs
- create the helper

Screenshot of docs:

image

### DIFF
--- a/styles/dropdown.css
+++ b/styles/dropdown.css
@@ -1,0 +1,124 @@
+/*---
+section: Helpers
+title: Dropdown
+---
+
+#### Variables
+```
+  dropdown-bg: '#fff'
+```
+
+#### Example
+```example:html
+<div class="dropdown open">
+  <button class="button dropdown-toggle">Dropdown Button</button>
+  <div class="dropdown-menu">
+    <button class="dropdown-item">Button item</button>
+    <button class="dropdown-item">Button item</button>
+    <a href="" class="dropdown-item">Link item</a>
+    <a href="" class="dropdown-item">Link item</a>
+  </div>
+</div>
+```
+*/
+.dropdown {
+  display: inline-block;
+  position: relative;
+}
+
+.dropdown-toggle {
+  transition: all ease .5s;
+
+  &::after {
+    display: inline-block;
+    width: 1em;
+    height: 1em;
+    margin-left: .25rem;
+    vertical-align: middle;
+    // triangle mixin here
+    content: '';
+  }
+
+  &:focus {
+    outline: 0;
+  }
+}
+
+.dropdown-menu {
+  display: none;
+  position: absolute;
+  top: 100%;
+  left: 0;
+  z-index: 1000;
+  float: left;
+  min-width: 160px;
+  margin: 2px 0 0;
+  border: var(--border-size) solid var(--color-border);
+  border-radius: var(--border-radius);
+  padding: 0;
+	background-color: var(--dropdown-bg);
+  background-clip: padding-box;
+  font-size: var(--font-size-body);
+  text-align: left;
+  list-style: none;
+}
+
+.dropdown-item {
+  box-sizing: border-box;
+  display: block;
+  width: 100%;
+  padding: calc(var(--spacing-xs) / 2) var(--spacing-xs);
+  clear: both;
+  font-size: inherit;
+  font-family: inherit;
+  font-weight: normal;
+  line-height: var(--line-height);
+  color: var(--color-link);
+  text-align: inherit;
+  white-space: nowrap;
+  background: none;
+  border: 0;
+
+  &:hover,
+  &:focus,
+  &:active {
+    color: var(--color-link-hover);
+    text-decoration: none;
+    background-color: var(--dropdown-bg);
+  }
+
+  &:active {
+    outline: 0;
+  }
+
+  &.disabled {
+    &:hover,
+    &:focus {
+      background-color: transparent;
+      background-image: none;
+      text-decoration: none;
+      cursor: not-allowed;
+    }
+  }
+}
+
+.open {
+  & > .dropdown-toggle::after {
+    // triangle mixin here
+    border: 1px solid black;
+  }
+
+  & > .dropdown-menu {
+    display: block;
+  }
+
+  & > a {
+    outline: 0;
+  }
+}
+
+.dropdown-menu-right {
+  right: 0;
+  left: auto;
+}
+
--- a/styles/dropdown.css
+++ b/styles/dropdown.css
@@ -12,9 +12,20 @@ title: Dropdown
 ```example:html
 <div class="dropdown open">
   <button class="button dropdown-toggle">Dropdown Button</button>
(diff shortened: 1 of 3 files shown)


styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+  top: 100%;
+  left: 0;
+  z-index: 1000;
+  float: left;
+  min-width: 160px;
+  margin: 2px 0 0;
+  border: var(--border-size) solid var(--color-border);
+  border-radius: var(--border-radius);
+  padding: 0;
+	background-color: var(--dropdown-bg);
```
jgallen23: weird indention here

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+    &:focus {
+      background-color: transparent;
+      background-image: none;
+      text-decoration: none;
+      cursor: not-allowed;
+    }
+  }
+}
+
+.open {
```
jgallen23: this might be a little too generic of a class name

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+.dropdown-toggle {
+  transition: all ease .5s;
+
+  &::after {
+    display: inline-block;
+    width: 1em;
+    height: 1em;
+    margin-left: .25rem;
+    vertical-align: middle;
+    // triangle mixin here
```
jgallen23: // isn't a valid comment

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+    color: var(--color-link-hover);
+    text-decoration: none;
+    background-color: var(--dropdown-bg);
+  }
+
+  &:active {
+    outline: 0;
+  }
+
+  &.disabled {
```
jgallen23: should this be dropdown-disabled?  should we add an example of this?

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+  min-width: 160px;
+  margin: 2px 0 0;
+  border: var(--border-size) solid var(--color-border);
+  border-radius: var(--border-radius);
+  padding: 0;
+	background-color: var(--dropdown-bg);
+  background-clip: padding-box;
+  font-size: var(--font-size-body);
+  text-align: left;
+  list-style: none;
```
jgallen23: should we use border-radius default on this menu?

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+  & > .dropdown-menu {
+    display: block;
+  }
+
+  & > a {
+    outline: 0;
+  }
+}
+
+.dropdown-menu-right {
```
jgallen23: what is this? should we add example?

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+      background-image: none;
+      text-decoration: none;
+      cursor: not-allowed;
+    }
+  }
+}
+
+.open {
+  & > .dropdown-toggle::after {
+    // triangle mixin here
```
jgallen23: not valid comment

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+    border: 1px solid black;
+  }
+
+  & > .dropdown-menu {
+    display: block;
+  }
+
+  & > a {
+    outline: 0;
```
jgallen23: what it this for?

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+      text-decoration: none;
+      cursor: not-allowed;
+    }
+  }
+}
+
+.open {
+  & > .dropdown-toggle::after {
+    // triangle mixin here
+    border: 1px solid black;
```
jgallen23: colors should be variables

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+  padding: 0;
+	background-color: var(--dropdown-bg);
+  background-clip: padding-box;
+  font-size: var(--font-size-body);
+  text-align: left;
+  list-style: none;
+}
+
+.dropdown-item {
+  box-sizing: border-box;
```
jgallen23: I think we have this on all element

styles/dropdown.css
```diff
@@ -0,0 +1,124 @@
...
+  display: block;
+  width: 100%;
+  padding: calc(var(--spacing-xs) / 2) var(--spacing-xs);
+  clear: both;
+  font-size: inherit;
```
jgallen23: how come these need to be set?

jgallen23:
should we have a hover style for this?

chrisopedia:
There's hover states for the menu items but given you can use a button or anything else I'd leave the hover state of that to whatever component triggers the dropdown.

styles/dropdown.css
chrisopedia: keeps things consistent in case someone uses a `<h* />` or something like that.

styles/dropdown.css
chrisopedia: Since there's no properties set on it and its all parent class there's no harm in using this.

styles/dropdown.css
chrisopedia: we can add an example; I don't see the need to overspecify the disabled class; rather prefer to keep the state type classes simple. But up to you.

styles/dropdown.css
chrisopedia: Dont' see a reason not, but if we have a better design that doesn't use it, I'm fine with that.

styles/dropdown.css
chrisopedia: Puts the menu to the right side.

styles/dropdown.css
jgallen23: this just seems risky to me. @dawnerd thoughts?

styles/dropdown.css
jgallen23: but what if we have an `.open` class that adds other things on the site?

styles/dropdown.css
jgallen23: ping

styles/dropdown.css
chrisopedia: classes as generic as that (ie .`open`, `closed`) and whatnot should be left for states and since we control that, its just a matter of paying attention.  I think this works and if we run into that issue, its a simple change, but no need to overcomplicate things until then. #my2cents

styles/dropdown.css
chrisopedia: zeros out outline which (should is normally set to focus on tab selection for accessibility). We could remove this if you want.

styles/dropdown.css
chrisopedia: This enforces in cases where something tries to override it.

styles/dropdown.css
jgallen23: it's not really consistent with the other stuff we've done though

chrisopedia: how so?

PR #105, opened 2016-09-03
Create triangle mixin and helper

chrisopedia:

1. we use triangles for everything
2. can be used for dropdown toggles

- create mixin as js for directional logic
- create helper to showcase
- create mdcss

Screenshot of docs:

image

### DIFF
--- a/styles/helpers/triangle.css
+++ b/styles/helpers/triangle.css
@@ -1,0 +1,25 @@
+/*---
+section: Helpers
+title: Triangle
+---
+
+```example:html
+<b class="triangle-up"></b>
+<b class="triangle-right"></b>
+<b class="triangle-down"></b>
+<b class="triangle-left"></b>
+```
+*/
+
+.triangle-up {
+@mixin triangle 10px, up;
+}
+.triangle-right {
+@mixin triangle 10px, right;
+}
+.triangle-down {
+@mixin triangle 10px;
+}
+.triangle-left {
+@mixin triangle 10px, left;
+}
--- a/styles/mixins/triangle.js
+++ b/styles/mixins/triangle.js
@@ -1,0 +1,48 @@
+'use strict';
+
+module.exports = function( config ) {
+  return function( mixin, size, direction ) {
+    if ( ! size ) {
+      size = '4px';
+    }
+    let borders;
+    const styles = {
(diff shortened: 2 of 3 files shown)


styles/helpers/triangle.css
```diff
@@ -0,0 +1,25 @@
...
+```example:html
+<b class="triangle-up"></b>
+<b class="triangle-right"></b>
+<b class="triangle-down"></b>
+<b class="triangle-left"></b>
+```
+*/
+
+.triangle-up {
+@mixin triangle 10px, up;
```
jgallen23: minor, but should these be tabbed in?

styles/mixins/triangle.js
```diff
@@ -0,0 +1,48 @@
+'use strict';
+
+module.exports = function( config ) {
```
jgallen23: are we able to make a triangle like this with this mixin?

screenshot 2016-09-05 at 8 44 49 pm

chrisopedia: No, but something like that, I'd rather use an SVG so you can scale it and keep it clean.  But we could add support for it I guess if we need to.

jgallen23: just curious, but why svg?

chrisopedia: Its an icon, so it seems like a perfect use case.  Requires 1 element, scales with the font, changes color with the parent, easier to style (doesn't require 2 selectors).

styles/helpers/triangle.css
chrisopedia: probably.

styles/mixins/triangle.js
jgallen23: I don't follow.  What do you mean its an icon?  Should just be an `:after` on the selected tab state

chrisopedia: Oh.  You cut off the image then.  It looks like a chat bubble icon.

chrisopedia: We could still probably use CSS transforms to make this possible with the current mixin if need be.

chrisopedia: Or just use a right facing triangle and hide it behind the tab.  That would work.

PR #109, opened 2016-09-05
Create button sizes

chrisopedia:

1. we need different button sizes
2. currently only think we need small and large sizes

- converts border-radius config to a nested object for various sizes
- update the existing border helper to use new default var
- update the existing button mixin to allow passed in borderRadius for cases where we need to turn it off (ie women in the workplace/leanin)
- create button mixin for sizes to change padding/border-radius and font-size; adds as a modifier to existing `.button` (i.e. `.button .button-lg`)
- create `.button-lg` and `.button-sm` helpers with new mixin

Screenshot of docs:

image

### DIFF
--- a/conf/default-vars.yaml
+++ b/conf/default-vars.yaml
@@ -48,7 +48,10 @@ vars:
 
   border:
     size: 1px
-    radius: 3px
+    radius:
+      default: 3px
+      lg: 5px
+      sm: 2px
 
   #messages
   message:
--- a/conf/default-vars.yaml
+++ b/conf/default-vars.yaml
@@ -53,6 +53,12 @@ vars:
       lg: 5px
       sm: 2px
 
+  button:
+    radius:
+      default: 3px
+      lg: 5px
+      sm: 2px
+
   #messages
   message:
     border-radius: 2px
--- a/styles/helpers/buttons.css
+++ b/styles/helpers/buttons.css
@@ -14,7 +14,7 @@ title: Primary Button
 ```
 */
 .button {
-  @mixin button var(--color-button-primary-bg), var(--color-button-primary-color), transparent, var(--border-radius-default), color(var(--color-button-primary-bg) blackness(20%)), var(--color-button-primary-color);
+  @mixin button var(--color-button-primary-bg), var(--color-button-primary-color), transparent, var(--button-radius-default), color(var(--color-button-primary-bg) blackness(20%)), var(--color-button-primary-color);
 }
 
 /*---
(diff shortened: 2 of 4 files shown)


conf/default-vars.yaml
```diff
@@ -48,7 +48,10 @@ vars:
 
   border:
     size: 1px
-    radius: 3px
+    radius:
+      default: 3px
+      lg: 5px
```
jgallen23: these variables should be in button, not border

styles/helpers/buttons.css
```diff
@@ -27,5 +27,22 @@ title: Outline Button
 ```
 */
 .button-outline {
-  @mixin button transparent, var(--color-button-primary-bg), var(--color-button-primary-bg), var(--color-button-primary-bg), var(--color-button-primary-color);
+  @mixin button transparent, var(--color-button-primary-bg), var(--color-button-primary-bg), var(--border-radius-default), var(--color-button-primary-bg), var(--color-button-primary-color);
+}
+
+/*---
+section: Buttons
+title: Button sizes
```
jgallen23: so is a size required now too?

styles/helpers/buttons.css
```diff
@@ -27,5 +27,22 @@ title: Outline Button
...
+---
+```example:html
+<button class="button button-lg">Large Button</button>
```
jgallen23: does this work with outline buttons too?

conf/default-vars.yaml
chrisopedia: any reason why?  Don't you want your border-radius to be consistent across a project?

styles/helpers/buttons.css
chrisopedia: no, its just a modifier.

styles/helpers/buttons.css
chrisopedia: I don't see why not.  Its just a modifier.

conf/default-vars.yaml
jgallen23: it seems odd to me to require modules with a border to have the same exact border radius as buttons. Just seems like unnecessary coupling between the two.  could be swayed though

styles/helpers/buttons.css
jgallen23: can we put that in the example?

conf/default-vars.yaml
chrisopedia: I would argue that if you're using border-radius on any element, design-wise you'd want them all to match for consistency.  I'd imagine its more the exception to have something with drastically different border-radius.

styles/mixins/button.css
```diff
@@ -34,14 +34,14 @@ Outputs
 }
 ```
 */
-@define-mixin button $background, $color, $border, $hoverBackground, $hoverColor {
+@define-mixin button $background, $color, $border, $borderRadius, $hoverBackground, $hoverColor {
```
jgallen23: need to update the docs above. you can remove output, just want the arguments for the mixin

PR #121, opened 2016-09-12
Update to include flexbox and add docs

chrisopedia:

1. adds documentation for flexbox
2. gives a simple example

- add in docs
- add to output

Screenshot of docs:

image

### DIFF
--- a/styles/helpers/flex-grid.css
+++ b/styles/helpers/flex-grid.css
@@ -1,3 +1,64 @@
+/*---
+section: Helpers
+title: Flexbox Grid
+---
+
+#### Variables
+```
+  grid-columns: 12
+  grid-gutters: 15px
+```
+
+#### Breakpoints
+```
+  desktop-wide:
+    max-width: 100vw
+    min-width: 1440px
+    content: 1440px
+    col: xl
+    default: '{{!core.mobileFirst}}'
+    bleed: 1600px
+  desktop:
+    max-width: 1439px
+    min-width: 1024px
+    content: 1024px
+    col: lg
+  tablet:
+    max-width: 1023px
+    min-width: 768px
+    content: 768px
+    col: md
+  mobile:
+    max-width: 767px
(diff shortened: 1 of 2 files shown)


styles/helpers/flex-grid.css
```diff
@@ -1,3 +1,64 @@
...
+
+#### Variables
+```
+  grid-columns: 12
+  grid-gutters: 15px
+```
+
+#### Breakpoints
+```
+  desktop-wide:
```
jgallen23: these are all dynamic, I don't think we should include the values

styles/helpers/flex-grid.css
```diff
@@ -1,3 +1,64 @@
...
+ | --- | --- | --- | --- | --- | --- |
+ | Container width | None (auto) | 320px - 767px  | 768px - 1023px | 1024px - 1439px | 1440px - 100vw |
+ | Class prefix | `.flex-` | `.flex-sm-` | `.flex-md-` | `.flex-lg-` | `.flex-xl-` |
+ | # of columns | 12 | 12 | 12 | 12 | 12 |
+ | Gutter width | 15px on each side | 15px on each side | 15px on each side | 15px on each side | 15px on each side |
+ | Nestable | Yes | Yes | Yes | Yes | Yes |
+ | Offsets | Yes | Yes | Yes | Yes | Yes |
+ | Column ordering | Yes | Yes | Yes | Yes | Yes |
+
+*/
```
jgallen23: do we have a way to order or stretch or any of that other fun stuff?

chrisopedia: You have the same classes as the grid, to do `push`, `pull` and I've added as much as I can figure out from the distro file.  What is in the docs seems to have split every single flex property into a separate class which makes it hard for me to figure out what is supposed to be combined to make a lot of the magic you want.  Might be better to have the author of this write the docs.

PR #124, opened 2016-09-14
shows file size in bytes for output files

orthagonal:


tasks/css.js
```diff
@@ -189,8 +190,8 @@ class CssTask {
     const output = path.join(this.config.core.dist, outputName);
     fs.writeFileSync(output, this.result.css);
     fs.writeFileSync(`${output}.map`, this.result.map);
-    log(`Wrote: ${this.input} → ${output}`);
-    log(`Wrote: ${this.input}.map → ${output}.map`);
+    log(`Wrote: ${this.input} → ${output} (${bytesize.stringSize(this.result.css, true)}), `);
+    log(`Wrote: ${this.input}.map → ${output}.map (${bytesize.stringSize(this.result.map, true)})`);
```
jgallen23: don't need filesize for map

PR #125, opened 2016-09-16
Create; mobile only option for menu slide

chrisopedia:

1. Adds another modifier to allow users to only have hamburger menu on
mobile rather than on desktop.
2. Currently on some clients we're having to override a bunch of the desktop CSS because the default is to have the hamburger menu show

- create `mobile-only` option
- add docs

Screenshot of docs:

image


styles/helpers/menu-slide.css
```diff
@@ -42,6 +42,26 @@ title: Menu Slide
 </div>
 <div class="col-11"><h1>ClientKit</h1></div>
 ```
+
+### Mobile Only
+
+```example:html
+<div class="col-9"><h1>ClientKit</h1></div>
+<div class="col-3">
+  <div class="menu-slide mobile-only">
```
jgallen23: what if I want tablet too?

jgallen23: I don't like the `mobile-only` class. It's confusing because the stuff underneath still shows up on non-mobile

styles/mixins/menu-slide.css
```diff
@@ -102,4 +102,38 @@ title: Mobile Menu
...
+      }
+      &:checked {
+        & ~ .menu-container {
+          right: 60%;
+          overflow-y: auto;
+        }
+      }
+      @media (--tablet) {
+        & ~ .menu-container {
+          position: fixed;
```
jgallen23: this is the exact same code as above, we should optimize this

styles/mixins/menu-slide.css
```diff
@@ -102,4 +102,38 @@ title: Mobile Menu
       }
     }
   }
+  &.mobile-only {
+    & .menu-btn {
+      & + label {
+        display: none;
+
+        @media (--tablet) {
```
jgallen23: what does this do?

styles/mixins/menu-slide.css
```diff
@@ -102,4 +102,38 @@ title: Mobile Menu
       }
     }
   }
+  &.mobile-only {
```
jgallen23: this says mobile only, but I only see tablet media queries

styles/helpers/menu-slide.css
chrisopedia: k.  What would you prefer?

styles/mixins/menu-slide.css
chrisopedia: I tired, but couldn't figure out how to.  Don't think we have extends options in PostCSS, and I need this nested underneath the parent selector.

styles/mixins/menu-slide.css
chrisopedia: Hides the hamburger menu on desktop, but shows up on tablet down.

styles/mixins/menu-slide.css
chrisopedia: Yeah, We only need tablet down.

jgallen23: if it's `mobileFirst: true`, this will be tablet up

styles/helpers/menu-slide.css
chrisopedia: I chose `tablet-down` since that matches the media query.

styles/mixins/menu-slide.css
chrisopedia: Its now using a newer custom media query ensures that using `tablet-down` will force it regardless of media query approach.

styles/mixins/menu-slide.css
chrisopedia: We'd need to use something like [davidtheclark/postcss-simple-extend](https://github.com/davidtheclark/postcss-simple-extend) to fix that.  Not sure we want to add this for just this one use case.

chrisopedia:
@jgallen23 submitted some new fixes for this.

PR #127, opened 2016-09-18
Create; custom media queries for each breakpoint

chrisopedia:

So we can target breakpoints regardless of mobile first or desktop first approach

- There is nothing that is actually output since these are just media queries in the `:root{}` selector, so I've created docs to show how to use them.
- I feel like there is probably a better way to do this using the JS mixin style but after exploring some options, I couldn't figure out how to accomplish it and this seems to be the simplest approach by just adding in the `:root{}` media queries.  I tested trying to use variables in the custom media queries and that didn't work either.  Don't know if thats a limitation of PostCSS or what.

I looked at using [postcss/postcss-media-minmax](https://github.com/postcss/postcss-media-minmax), but seemed like overkill.  Same goes for [postcss/postcss-custom-media](https://github.com/postcss/postcss-custom-media)

Screenshot of docs:

image

### DIFF
--- a/styles/helpers/custom-media-queries.css
+++ b/styles/helpers/custom-media-queries.css
@@ -1,0 +1,69 @@
+/*---
+section: Helpers
+title: Custom Media Queries
+---
+
+#### Usage
+
+For all of the breakpoints there are 3 options.  An `up` to target from a min-width of a particular breakpoint and up.  A `down` to target from a max-width of a particular breakpoint and down.  And an `only` to target from a min-width of a breakpoint to the max-width of that same breakpoint.
+
+#### Up
+
+```example:css
+@media (--tablet-up) {
+  body {
+    background-color: red;
+  }
+}
+```
+
+will compile to:
+
+```example:css
+@media (min-width: 768px) {
+  body {
+    background-color: red;
+  }
+}
+```
+
+#### Down
+
+```example:css
+@media (--tablet-down) {
+  body {
+    background-color: green;
+  }
+}
+```
+
+will compile to:
+
+```example:css
+@media (max-width: 1023px) {
+  body {
+    background-color: red;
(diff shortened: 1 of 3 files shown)


styles/helpers/mediaqueries.css
```diff
@@ -1,3 +1,19 @@
 :root {
+  @custom-media --desktop-wide-up (min-width: 1440px);
```
jgallen23: we need to use a js mixin so it can access the config values

styles/helpers/custom-media-queries.css
```diff
@@ -0,0 +1,69 @@
+/*---
+section: Helpers
```
jgallen23: why is this a separate file?

jgallen23:
FYI - postcss-custom-media is already included in clientkit

styles/helpers/custom-media-queries.css
chrisopedia: So I could create the documentation for how to use it.  Seems like a good place to put it.  Did you have somewhere better in mind to place it?

styles/helpers/mediaqueries.css
chrisopedia: How would something like that work?  All of the JS mixins I've seen us create output some sort of CSS declaration that is wrapped in the selector when you use the mixin as in the helpers.

ie. 
```
@grid

# outputs

.col-12 {
  float: left;
  width: 100%;
}
```

This needs to just output the `@custom-media --desktop-up (min-width: var(--desktop-min)` type of thing where you'd have to pass in content.  In PostCSS mixins you have the `@mixin-content` for these kinds of things but how would you do that in JS?

jgallen23:
closing. going to approach this differently

PR #131, opened 2016-09-21
Flex grid fixes

Belelros:

1. Because we don't have paddings (as we do on the normal grid).
2. Because we don't have a way to say `flex-md-row` right now.

- Add paddings to all columns
- Add flex-row selectors for breakpoints.


jgallen23:
we also probably need a flex-plain for no padding option, but this looks good for now

PR #132, opened 2016-09-23
Add new breakpoints

Belelros:
- Adding new breakpoints for `only` and `up` / `down` where appropriate.


tasks/css.js
```diff
@@ -69,6 +69,36 @@ class CssTask {
...
+      if (i !== bps.length < 1) {
+        this.customMedia[`${breakpoint}-down`] = `(max-width: ${breakpointObj.max})`;
+      }
+
+      // First one should be wide desktop so no up
+      if (i !== 0) {
+        this.customMedia[`${breakpoint}-up`] = `(min-width: ${breakpointObj.min})`;
+      }
+
+      this.cssVars[`breakpoint-${breakpoint}`] = width;
```
jgallen23: This is set above too. Do we still need it here?

Belelros: That's a leftover. I was intending to keep the same functionalities and group this new feature into this loop.

PR #134, opened 2016-09-25
bring in theme to clientkit

orthagonal:
--brings mdcss-theme-clientkit into styles/themes folder
--updated some of the template.ejs to make it work with lodash


package.json
```diff
@@ -51,9 +51,9 @@
     "lodash.capitalize": "^4.2.0",
     "lodash.debounce": "^4.0.6",
     "lodash.reduce": "^4.5.1",
+    "lodash.template": "^4.4.0",
```
jgallen23: lets just use all of lodash, probably easier at this point

styles/theme/assets/anchor.js
```diff
@@ -0,0 +1,343 @@
+/**
```
jgallen23: lets add to a styleguide folder instead of styles

PR #135, opened 2016-09-26
Mobile body sizes

Belelros:

1. Because it adds a way to add font-sizes for mobile across the site.

- Adding a new group for small font-sizes.
- Including them on the typography.css

### DIFF
--- a/styles/helpers/typography.css
+++ b/styles/helpers/typography.css
@@ -136,14 +136,26 @@ title: Size
 */
 .font-large {
   font-size: var(--font-size-large);
+
+  @media(--mobile) {
+    font-size: var(--font-small-size-large);
+  }
 }
 
 .font-small {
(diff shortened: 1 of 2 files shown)


styles/helpers/typography.css
```diff
@@ -136,14 +136,26 @@ title: Size
...
+  @media(--mobile) {
+    font-size: var(--font-small-size-large);
+  }
 }
 
 .font-small {
   font-size: var(--font-size-small);
+
+  @media(--mobile) {
+    font-size: var(--font-small-size-small);
```
jgallen23: What about mobile instead of small? This reads a little funny

Belelros:
Did that just in case we decide to add a new one for tablet later on (medium). But I guess you are right!

jgallen23:
I think we would just do `font-tablet-size-small`

PR #137, opened 2016-10-03
Add mobile spacing

dawnerd:


dawnerd:
It should just work right? How does the breakpoint handler deal with the other cases like font sizes?

jgallen23:
oh, you are right!

jgallen23:
actually, mobile will match everything, so default will never get hit

dawnerd:
Hmm, sounds like we should update the handler then.

PR #139, opened 2016-10-03
Read Only Inputs

dawnerd:
Not sure what other styles we could add here...

### DIFF
--- a/styles/mixins/input.css
+++ b/styles/mixins/input.css
@@ -27,4 +27,7 @@ title: Input
     border-bottom: 1px dashed $borderDisabled;
     color: $colorDisabled;
   }
+  &:read-only {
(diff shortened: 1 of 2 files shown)


styles/mixins/input.css
```diff
@@ -27,4 +27,7 @@ title: Input
     border-bottom: 1px dashed $borderDisabled;
     color: $colorDisabled;
   }
+  &:read-only {
+    cursor: text;
```
jgallen23: should we be changing the text color or bg color?

dawnerd: I thought about that but it didn't make much sense for cases where we output a url for a user to copy. I did try using the disabled colors but that didn't really look right.

PR #142, opened 2016-10-12
Adding an injection for hashes

ecwillis:
1. It adds an injection for static files when using file hashes.

- injection script
- bug fix


lib/urlHashes.js
```diff
@@ -1,17 +1,19 @@
 'use strict';
-const hash = require('rev-hash');
```
jgallen23: If we are going to remove this, make sure to do it in package.json too

lib/injectHash.js
```diff
@@ -0,0 +1,27 @@
...
+const glob = require('glob');
+const fs = require('fs');
+
+module.exports = function(fileName, hashedName, conf) {
+  const nameArr = fileName.split('.');
+  const regex = new RegExp(`(\\<\\!--\\s{0,}clientkit:${nameArr.join('\\.')}\\s{0,}--\\>)([\\t\\n\\s\\<a-zA-Z0-9\\"\\/\\.\\>\\=-]*)(\\<\\!--\\s{0,}clientkit:end\\s{0,}-->)`);
+  const fileLink = `<link rel="stylesheet" href="/${conf.prefix}${hashedName}">`;
+  glob(conf.file, (err, files) => {
+    if (err) {
+      console.log('Error finding injection files');
```
jgallen23: should use logr like every thing else

lib/injectHash.js
```diff
@@ -0,0 +1,27 @@
...
+    }
+
+    files.forEach((file) => {
+      fs.readFile(file, (fileErr, content) => {
+        let contentString = content.toString();
+        contentString = contentString.replace(regex, `\$1\n${fileLink}\n\$3`);
+        fs.writeFile(file, contentString, (writeErr) => {
```
jgallen23: This writes the file every time. Should we check if it exists first?

lib/urlHashes.js
ecwillis: Done.

lib/injectHash.js
ecwillis: Fixed it so it write less. I am overwriting the original file so it uses debounce to most likely write it once.

lib/injectHash.js
```diff
@@ -0,0 +1,56 @@
...
+
+  glob(conf.files, (err, files) => {
+    if (err) {
+      log('Error finding injection files');
+      return;
+    }
+
+    files.forEach((file) => {
+      let contentString = cache[file];
+      if (!cache[file]) {
```
jgallen23: I'm a little concerned with this.  What if the view html changes?

lib/urlHashes.js
```diff
@@ -1,17 +1,12 @@
...
 let map = {};
 
 module.exports.hash = (inputName, inputContent) => {
-  let hashKey;
-  if (typeof inputContent === 'string') {
-    hashKey = hash(Buffer.from(inputContent));
-  } else {
-    hashKey = hash(inputContent);
-  }
+  const hashKey = crypto.createHash('md5').update(inputContent).digest('hex').slice(0, 10);
```
jgallen23: so will this work when inputContent is a string or buffer?

public/index.html
```diff
@@ -0,0 +1,13 @@
+<!DOCTYPE html>
```
jgallen23: can we move this into the test directory?

lib/injectHash.js
```diff
@@ -0,0 +1,56 @@
...
+    cli: {
+      lineColor: 'yellow'
+    }
+  }
+});
+
+const cache = {};
+
+const writeFile = debounce((file, contentString) => {
+  fs.writeFileSync(file, contentString);
```
jgallen23: now that I think about it, we are already deboucing higher up the funnel, so we might not need this

commands/run.js
```diff
@@ -1,17 +1,62 @@
 'use strict';
+const async = require('async');
+const _ = require('lodash');
```
jgallen23: lodash isn't in package.json

commands/run.js
```diff
@@ -1,17 +1,62 @@
...
   // Tasks
-  mkdirp.sync(config.core.dist);
-  if (config.stylesheets) {
-    Object.keys(config.stylesheets).forEach(style => cssProcessor.runTaskAndWrite(config, process.cwd(), style, config.stylesheets[style]));
-  }
-  if (config.scripts) {
-    Object.keys(config.scripts).forEach(script => jsProcessor(config, process.cwd(), script, config.scripts[script]));
-    lintProcessor(config, process.cwd());
-  }
+  async.auto({
```
jgallen23: will this work for the watchers too? `--env dev`

commands/run.js
```diff
@@ -1,17 +1,62 @@
 'use strict';
+const async = require('async');
+const _ = require('lodash');
 const mkdirp = require('mkdirp');
+const cssProcessor = require('../tasks/css.js');
```
jgallen23: these where inside runAll because they take forever to require

jgallen23:
```
07:21 $ ~/libraries/clientkit/app.js --env dev
  Using local config directory: /home/ubuntu/clients/cuddle-bright/splash/clientkit
  Changed: clientkit/default-colors.yaml
  Changed: clientkit/default-spacing.yaml
  Changed: clientkit/default-vars.yaml
  Changed: clientkit/default.yaml
  Changed: styles.css
  Processed ../../../libraries/clientkit/styles/default.css in 3.064 sec
  Wrote: dist/clientkit.css (64.54 Kb), 
(node:10743) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 1): TypeError: cb is not a function
  Processed styles.css in 5.103 sec
  Wrote: dist/styles.css (78.64 Kb), 
(node:10743) UnhandledPromiseRejectionWarning: Unhandled promise rejection (rejection id: 2): TypeError: cb is not a function
```

commands/run.js
ecwillis: Added

ecwillis:
I added some `noops` to the processors that will take care of the missing `cb` issue.

PR #144, opened 2016-10-25
Adds tablet breakpoint to fix font sizes in mobile-first

dawnerd:
Simple fix. When we switched to css mixins old functionality broke. Added a tablet breakpoint that'll reset the size back when in mobile first and shouldn't do anything on desktop first.


jgallen23:
closing this in favor of using mobile-only breakpoints

PR #147, opened 2016-11-01
Easing

mturnwall:

1. Better easing options for smoother animations

- New yaml config file for easing variables
- Examples page to demo what each easing option does

I added an example page to demonstrate how each easing function behaves. You can use a drop down to choose an easing function and alter the time of the animation.


examples/easing/index.html
```diff
@@ -0,0 +1,169 @@
+<!DOCTYPE html>
```
jgallen23: FYI, we don't use the example folder anymore.  The docs are inline with the CSS

PR #148, opened 2016-11-05
Build timeout

dawnerd:
Seems to solve the issue. Couldn't find anywhere that was actually running the watchers more than they should be. Just a stupid race condition.


jgallen23:
Looks good

PR #152, opened 2016-11-17
refactor: js maps in tasks/scripts.js

orthagonal:


tasks/scripts.js
```diff
@@ -27,8 +28,11 @@ class ScriptsTask extends ClientKitTask {
 
     if (this.options.minify) {
       currentTransform = currentTransform.transform(uglifyify, { global: true });
+      const result = currentTransform
```
jgallen23: shouldn't this be outside of the minify check?

jgallen23:
lets add a sourcemap option and default it to true

PR #154, opened 2016-11-17
Css maps when 'sourcemap' is not false

orthagonal:


tasks/css.js
```diff
@@ -187,26 +187,34 @@ class CSSTask extends ClientKitTask {
     if (this.options.minify) {
       processes.push(cssnano());
     }
-    fs.readFile(input, (readErr, buf) => {
```
jgallen23: need to fix this

tasks/css.js
```diff
@@ -200,7 +200,12 @@ class CSSTask extends ClientKitTask {
               }
             });
           }
-
+          // write the source map if indicated:
+          if (this.options.minify && this.options.sourcemap !== false) {
```
jgallen23: why are we just writing this if minify?

tasks/css.js
```diff
@@ -200,7 +200,12 @@ class CSSTask extends ClientKitTask {
...
+            return this.write(`${outputFilename}.map`, JSON.stringify(result.map), () => {
```
jgallen23: can we optimize this by using async.parallel?

jgallen23: or use async auto and do the sourcemaps check inside the fn

tasks/css.js
```diff
@@ -187,28 +187,44 @@ class CSSTask extends ClientKitTask {
...
+            }
+          });
+        }
+        done();
+      }],
+      sourcemaps: ['postcss', (results, done) => {
+        // write the source map if indicated:
+        if (results.postcss.map && this.options.sourcemap !== false) {
+          this.write(`${outputFilename}.map`, JSON.stringify(results.postcss.map), done);
+        }
```
jgallen23: Need to call done if false

tasks/css.js
```diff
@@ -187,28 +187,44 @@ class CSSTask extends ClientKitTask {
...
+      }]
+    }, (err, results) => {
+      if (err) {
+        this.log(['error'], err.stack);
+      }
+      this.write(outputFilename, results.postcss.css, callback);
```
jgallen23: Can we do the write as part of auto?

tasks/css.js
```diff
@@ -187,28 +187,44 @@ class CSSTask extends ClientKitTask {
...
-            });
-          }
-
-          this.write(outputFilename, result.css, callback);
-        }, (err) => {
-          if (err) {
-            this.log(['error'], err.stack);
+    async.auto({
+      contents: (done) => {
+        fs.readFile(input, (readErr, buf) => {
```
jgallen23: Just put done as second parameter.

jgallen23:
This is a lot cleaner. Nice job

PR #155, opened 2016-11-18
adds hash and initalize tasks, fixes task series

orthagonal:


app.js
```diff
@@ -47,13 +45,11 @@ const main = () => {
   }
   const task = argv._.length === 0 ? 'default' : argv._;
   log(`Running ${task}...`);
-  //create dist directory
-  mkdirp.sync(conf.dist);
   loadTasks(conf, (err, runner) => {
     if (err) {
       throw err;
     }
-    runner.run(task);
```
orthagonal: Fixing this seems to make things run in the right order.

conf/default.yaml
```diff
@@ -1,11 +1,12 @@
...
+minify: true
 tasks:
   default:
-    -
-      - 'eslint'
-      - 'stylesheets'
-      - 'scripts'
+    - 'initialize'
+    - 'eslint'
+    - 'stylesheets'
```
jgallen23: The middle 3 should be an array so they are parallel

tasks/hash.js
```diff
@@ -0,0 +1,25 @@
+'use strict';
+const async = require('async');
+const ClientKitTask = require('clientkit-task');
+const glob = require('glob');
+const hasher = require('../lib/urlHashes.js');
```
jgallen23: Can we pull this logic into this file?

tasks/hash.js
```diff
@@ -0,0 +1,25 @@
...
+  process(input, filename, processDone) {
+    async.auto({
+      filesToHash: (done) => {
+        // get the list of matching files
+        glob(input, {}, done);
+      },
+      hashNames: ['filesToHash', (results, done) => {
+        const hashMap = {};
+        results.filesToHash.forEach((file) => {
+          fs.renameSync(file, hasher.hash(file, fs.readFileSync(file)));
```
jgallen23: We should never do sync calls.

tasks/initialize.js
```diff
@@ -0,0 +1,24 @@
...
+const ClientKitTask = require('clientkit-task');
+const mkdirp = require('mkdirp');
+const rmdir = require('rmdir');
+const fs = require('fs');
+
+class InitTask extends ClientKitTask {
+  process(distDir, whatever, processDone) {
+    async.auto({
+      clearDist: (done) => {
+        if (fs.existsSync(distDir)) {
```
jgallen23: Exists should be a separate call that isn't sync

PR #157, opened 2016-11-19
Refactor help task

orthagonal:


.gitignore
```diff
@@ -1,5 +1,6 @@
 node_modules
 .idea
+dist
```
jgallen23: I want to keep checking this in, so we can see the output changes

conf/default-eslint.yaml
```diff
@@ -1,6 +1,7 @@
 tasks:
   eslint: '{{CKDIR}}/tasks/eslint.js'
 eslint:
+  description: 'Runs the indicated eslint config against the files you listed, and reports the results '
```
jgallen23: This should be on the class itself, not the options

conf/default.yaml
```diff
@@ -9,5 +9,6 @@ tasks:
       - 'stylesheets'
       - 'scripts'
     - 'hash'
+    - 'help'
```
jgallen23: Help shouldn't be shown by defaukt

tasks/help.js
```diff
@@ -0,0 +1,69 @@
+'use strict';
+const ClientKitTask = require('clientkit-task');
+const Logr = require('logr');
+class HelpTask extends ClientKitTask {
+
+  constructor(name, config, runner) {
+    config.log = new Logr({
```
jgallen23: What is this for?

jgallen23: This is already part of the class

tasks/help.js
```diff
@@ -0,0 +1,69 @@
...
+        }
+      }
+    });
+    super(name, config, runner);
+    this.calledOnce = false;
+  }
+
+  process(input, filename, processDone) {
+    if (this.calledOnce) {
+      return processDone();
```
jgallen23: We should override execute and not process. Process won't get called because it doesn't have any files or items

conf/default-help.yaml
```diff
@@ -0,0 +1,8 @@
+tasks:
+  help: '{{CKDIR}}/tasks/help'
+help:
+  active: true
```
jgallen23: Tasks already have a built in enabled property

conf/default-help.yaml
```diff
@@ -0,0 +1,8 @@
...
+  description: 'Prints various help info about your tasks'
+  logColor: 'white'
+  files:
```
jgallen23: Can remove this

tasks/help.js
orthagonal: I thought we might want help to have some special custom formatting options here, to make the help guide stand out better from the other tasks, but it's not necessary.

jgallen23: I would imagine help is only run as standalone, so you don't really need it to stand out

conf/default-help.yaml
```diff
@@ -0,0 +1,5 @@
+tasks:
+  help: '{{CKDIR}}/tasks/help'
+help:
+  enabled: true
```
jgallen23: Don't need to set this to true

conf/default-initialize.yaml
```diff
@@ -1,5 +1,6 @@
 tasks:
   initialize: '{{CKDIR}}/tasks/initialize.js'
 initialize:
+  description: 'Initializes the output directory so that everything will be available'
```
jgallen23: No description here

tasks/eslint.js
```diff
@@ -4,6 +4,10 @@ const CLIEngine = require('eslint').CLIEngine;
 const ClientKitTask = require('clientkit-task');
 
 class EslintTask extends ClientKitTask {
+  constructor(name, config, runner) {
+    super(name, config, runner);
```
jgallen23: Can we just set a property for the without needing a constructor?  Worst case a description is a function

conf/default.yaml
```diff
@@ -11,3 +11,4 @@ tasks:
     - 'hash'
   dev:
     - 'watcher'
+    - 'help'
```
jgallen23: help won't get called here since watcher stays open, you can remove

tasks/watcher.js
```diff
@@ -6,6 +6,11 @@ const debounce = require('lodash.debounce');
 const pathLib = require('path');
 
 class WatcherTask extends ClientKitTask {
+
+  get description() {
+    return 'In dev mode, this task will recompile your whole project when you make changes to the indicated files ';
```
jgallen23: remove `in dev mode` since that's configurable

PR #164, opened 2016-12-07
allows watches specified with {task, ignore}  in config

orthagonal:


tasks/watcher.js
```diff
@@ -4,16 +4,22 @@ const ClientKitTask = require('clientkit-task');
...
 
   process(tasks, watch, done) {
-    const ignored = new RegExp(this.options.ignore.join('|'));
-    this.log(`Ignoring: ${ignored}`);
+    // add together the top-level watcher 'ignore' expressions with this watch's specific 'ignore' expressions:
+    const files = this.options.files;
+    let ignoreString = this.options.ignore.join('|');
+    Object.keys(files).forEach((pathRoot) => {
+      if (typeof files[pathRoot] === 'object' && files[pathRoot].ignore) {
+        ignoreString = [ignoreString, files[pathRoot].ignore].join('|');
```
jgallen23: can you just do a this.options.ignore.concat(files[pathRoot].ignore)?  that doesn't affect the main object, right?

tasks/watcher.js
```diff
@@ -4,16 +4,22 @@ const ClientKitTask = require('clientkit-task');
...
-
 class WatcherTask extends ClientKitTask {
 
   get description() {
     return 'This task watches the indicated files for changes, and re-runs the other registered ClientKitTasks when an edit is made to them.';
   }
 
   process(tasks, watch, done) {
-    const ignored = new RegExp(this.options.ignore.join('|'));
-    this.log(`Ignoring: ${ignored}`);
```
jgallen23: we lost this

PR #166, opened 2016-12-20
CKDIR is process.cwd()

orthagonal:


lib/config.js
```diff
@@ -21,7 +20,7 @@ module.exports = (clientkitConfPath, projectConfPath, env, allDone) => {
           }
         ],
         context: {
-          CKDIR: path.join(__dirname, '..'),
+          CKDIR: process.cwd(),
```
jgallen23: This won't work  clientkit isn't in cwd when it's installed

lib/config.js
```diff
@@ -21,7 +21,7 @@ module.exports = (clientkitConfPath, projectConfPath, env, allDone) => {
           }
         ],
         context: {
-          CKDIR: path.join(__dirname, '..'),
+          CKDIR: path.relative(process.cwd(), '..'),
```
jgallen23: we _can't_ use process.cwd() for CKDIR

PR #172, opened 2016-12-21
wrapper to lazy-load and cache task modules

orthagonal:


lib/load-tasks.js
```diff
@@ -1,18 +1,30 @@
...
     if (Array.isArray(task)) { //task maps to other tasks
       taskFn = task;
-    } else {
-      const Cls = require(task);
-      //if task is config, pass the whole thing, little bit of a hack
-      const taskConfig = (config[name] && config[name].needsEntireConfig) ? config : config[name];
-      taskFn = new Cls(name, taskConfig, runner, log);
+    } else { // task is wrapped by a function that will lazy-load/cache it:
+      taskFn = (callback) => {
+        if (cachedTasks[name]) {
```
jgallen23: do a if !cachedTask and then set the value and then always do cachedTask[].execute(callback).

PR #173, opened 2016-12-23
use clientkit-core

orthagonal:


conf/default-hash.yaml
```diff
@@ -1,5 +1,5 @@
 tasks:
-  hash: '{{CKDIR}}/tasks/hash.js'
+  hash: 'clientkit-hash.js'
```
jgallen23: is the .js supposed to be there?

conf/default-stylesheets.yaml
```diff
@@ -1,9 +1,10 @@
 tasks:
-  stylesheets: '{{CKDIR}}/tasks/css.js'
+  stylesheets: 'clientkit-css'
 
 stylesheets:
   logColor: 'green'
-  assetPath: '{{assetPath}}'
+  globalMixins: '{{CKDIR}}/styles/mixins'
+  assetPath: '{{CKDIR}}/styles'
```
jgallen23: we should not change the assetPath

conf/default-help.yaml
```diff
@@ -1,4 +1,4 @@
 tasks:
-  help: '{{CKDIR}}/tasks/help'
+  help: 'help'
```
jgallen23: why did these change?

PR #179, opened 2017-07-11
Ripout docs

chrisopedia:

1. favor separate repo for docs

- rip out all inline docs
- move the dropdown helper to the helper directory

Notify or mention any users: @firstandthird/clientkit 

Fixes: [rip out inline docs on clientkit](https://app.asana.com/0/114313002471856/380102114018235)


demo/common.css
```diff
@@ -0,0 +1 @@
+@import '../styles/clientkit.css';
```
jgallen23: did `clientkit.css` not work?

chrisopedia: Nope.  It couldn't find anything.

PR #180, opened 2017-07-11
Update; change flex to flex-container

iamnewton:

1. confusion around the flex grid

- change `.flex` to `.flex-container`

Notify or mention any users: @firstandthird/clientkit 

Fixes: [Issue with flex row](https://app.asana.com/0/114313002471856/219282599651737)

### DIFF
--- a/styles/helpers/flex-grid.css
+++ b/styles/helpers/flex-grid.css
@@ -1,4 +1,4 @@
-.flex {
+.flex-container {
   flex: 1;
 }
 


chrisopedia:
```css
flex {
    flex: 1;
}
```

Thats way too generic of a class IMO.  I'm fine with whatever suggestions, but you can't just use `display: flex` to fix the original issue.

Belelros:
It just means behave as a flex-item should. I borrowed it from here: https://material.angularjs.org/latest/layout/children .

My comment around `display: flex` is that a flex-container (as your class suggest) is actually an element that has either `display: flex` or `display: inline-flex` but not an element with `flex: 1` because it's a property for flex-items, no containers. My suggestion was that  adding a `.flex-container` class with `display: flex` would be a consistency break since we already have `.display-{block,inline-block,inline}` helpers but lacking flex ones.

Belelros:
In order to fix the issue that Greg mentioned we should add the `display-flex` and `display-inline-flex` classes as well as demote some classes from the grid such as `flex-wrap` which is currently attached to the grid.

Grid concept requires gutters and negative margins yet Greg wants columns of the same width without having to use a grid at all.

PR #189, opened 2017-12-04
Flex fixes

AlejandroPerezMartin:

1. Adds support to flex baseline alignment properties
2. Adds flex-basis fallback for IE

- Add missing flex baseline aligners
- Adds flex-basis fallback for IE
- Display flex helpers

### DIFF
--- a/styles/mixins/flex-grid.js
+++ b/styles/mixins/flex-grid.js
@@ -124,6 +124,7 @@ module.exports = function (config) {
 
       for (let i = 1; i <= cols; i++) {
         block[`.${prefix}-${i}`] = {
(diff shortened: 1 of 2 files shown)


styles/mixins/flex-grid.js
```diff
@@ -7,13 +7,19 @@ module.exports = function (config) {
     'right': 'flex-end',
     'center': 'center',
     'space-around': 'space-around',
-    'space-between': 'space-between'
+    'space-between': 'space-between',
+    'baseline': 'baseline',
+    'first-baseline': 'first-baseline',
```
Antonio-Laguna: I think we can leave these two out. Other than that, looks great.

PR #191, opened 2018-01-03
3.0.0

Antonio-Laguna:


conf/default-vars.yaml
```diff
@@ -12,7 +12,6 @@ stylesheets:
         regular: 400
         semibold: 600
```
jgallen23: I feel like we should never have two different bold values defined.  I don't think we'll ever have a site that truly uses 4 weights - probably just 3

styles/helpers/typography.css
```diff
@@ -96,26 +96,22 @@ body {
   }
 }
 
-
 .heading--light {
```
jgallen23: do we want to keep the `--`?

styles/helpers/typography.css
```diff
@@ -96,26 +96,22 @@ body {
...
 .text-color-body {
   color: var(--color-body);
 }
+
 .text-color-heading {
   color: var(--color-heading);
 }
+
 .text-color-light {
   color: var(--color-heading-light);
```
jgallen23: this is a little misleading of a helper

conf/default-vars.yaml
Antonio-Laguna: Ok, we can always add it back.

styles/helpers/typography.css
Antonio-Laguna: I guess we can even remove this since just below we have `.text-color-light` which does exactly the same.

styles/helpers/typography.css
Antonio-Laguna: What do you mean?

jgallen23: the class name is text color light, but the variable is color heading light

test/expected/full.css
```diff
@@ -843,8 +843,8 @@ body {
 }
 .flex-1 {
   width: 8.333333333333334%;
-  -ms-flex-preferred-size: 8.333333333333334%;
-      flex-basis: 8.333333333333334%;
+  -ms-flex-preferred-size: auto;
+      flex-basis: auto;
```
jgallen23: is this backwards compatible?

test/expected/slim.css
```diff
@@ -3584,10 +3581,6 @@ hr {
         -ms-flex-direction: column;
             flex-direction: column;
   }
-.flex-row-sm-wrap {
```
jgallen23: do we use this anywhere?

Antonio-Laguna: Yeah we do but as this is a breaking release, better now than later. We don't rely on the fact that it adds the negative margin though. But we can't use `display-flex` in conjunction with that right now since it makes things get negative margins

test/expected/full.css
Antonio-Laguna: Yes, that's what we do to fix IE11, flex basis is just another way to set the width

test/expected/slim.css
jgallen23: hah - it was a breaking release - we are using it in production hah

Antonio-Laguna: It should only be a matter of replacing those classes:

`flex-row-wrap` -> `flex-wrap`
`flex-row-sm-wrap` -> `flex-sm-wrap`

I could maintain the name and remove the side effects, seemed misleading though.

PR #200, opened 2018-07-12
Add styles for floating label

AlejandroPerezMartin:
This is how a form looks in the styleguide with the default values:

screen shot 2018-07-12 at 13 02 40


Antonio-Laguna:
Please use another class which is more specific, `.floating-label` for example. `.form-group` is not implicit enough.

PR #242, opened 2021-04-23
Webpack 5

AlejandroPerezMartin:
Tried to update Clientkit to webpack 5, couldn't finish this because of some incompatibilities with some libraries


index.js
```diff
@@ -2,6 +2,7 @@
 /* eslint-disable no-console */
 const { fork } = require('child_process');
 const path = require('path');
+process.traceDeprecation = true;
```
AlejandroPerezMartin, line 5: to be removed

